PR #24, opened 2023-04-04
:recycle: - Move formik to formio-renderer.

### LINKED ISSUE #23: Move Formik to formio renderer

svenvandescheur:
closes #23


svenvandescheur:
@sergei-maertens FYI creating a lockfile from my machine results in an unusable file due to addresses pointing to an on premise npm repository. Can you re-create/add yarn.lock?

src/lib/renderer/renderer.tsx
```diff
@@ -65,22 +73,59 @@ export interface IRenderFormProps {
...
+ * @external {SubmissionContext} Provides `SubmissionContext`.
  */
-export const RenderForm = ({form}: IRenderFormProps): React.ReactElement => {
-  const children =
+export const RenderForm = ({
+  children,
+  configuration,
+  form,
+  formAttrs,
+  initialValues,
```
sergei-maertens, line 85: this doesn't make a lot of sense since `initialValues` is in essence equal to your `submission.data` - given an existing submission with data being loaded (or default values), that is the starting point for formik. There is no situation where you provide a submission with data AND initial values that are different.

src/lib/renderer/renderer.tsx
```diff
@@ -171,7 +216,7 @@ export const useComponentType = (
 
 /**
  * Custom hook resolving the value from `ValuesContext`.
- * @external {RenderContext} Expects `RenderContext` to be available.
+ * @external {SubmissionContext} Expects `SubmissionContext` to be available.
  */
 export const useValue = (
```
sergei-maertens, line 221: this approach is now obsolete because we don't have to fetch the value from submission.data anymore - it should be provided to formik `initialValues`, after which you can `const {values} = useFormikContext()` to obtain the form values and then `get(values, component.key)` to get the value of one particular component.

Alternatively, you can also do `const [{value}] = useField(component.key);` to get the value of the particular component from the formik state (I think this has my preference).

sergei-maertens (changes requested):

src/lib/renderer/renderer.tsx
```diff
@@ -65,29 +58,51 @@ export interface IRenderFormProps {
...
     form.components?.map((component: IRendererComponent, i: number) => (
       <RenderComponent key={component.id || i} component={component} />
     )) || null;
-  return <React.Fragment>{children}</React.Fragment>;
+
+  return (
+    <Formik initialValues={initialValues} onSubmit={onSubmit}>
+      {props => {
+        return (
+          <form onSubmit={props.handleSubmit} {...formAttrs}>
```
sergei-maertens, line 80:
```suggestion
      {({handleSubmit}) => {
        return (
          <form onSubmit={handleSubmit} {...formAttrs}>
```
I find the `props` here confusing as it is convention that your own component props are named like that. The formik docs are inconsistent - I'm seeing examples with destructuring, `formikProps` and `props`.

I think I prefer destructuring in this case.

src/lib/renderer/renderer.tsx
```diff
@@ -111,14 +126,14 @@ export interface IRenderComponentProps {
...
  * @external {RenderContext} Expects `RenderContext` to be available.
- * @external {SubmissionContext} Expects `SubmissionContext` to be available.
  */
 export const RenderComponent = ({component}: IRenderComponentProps): React.ReactElement => {
-  const callbacks = useContext(CallbacksContext);
-  const value = useValue(component);
   const Component = useComponentType(component);
+  const field = useField(component.key || OF_MISSING_KEY);
+  const {value, onBlur, onChange} = field[0];
+  const callbacks = {onBlur, onChange};
```
sergei-maertens, line 136:
```suggestion
  const [{value, onBlur, onChange}] = useField(component.key || OF_MISSING_KEY);
  const callbacks = {onBlur, onChange};
```
The args `meta` and `helpers` are not relevant at the moment, but they might become.

IMO this makes it a bit more obvious that the callbacks originate from formik and are re-wrapped for our own purposes.

I'm also not sure if we should pass `callbacks` as an object to the component, or just the callbacks themselves, but that'll become clearer later on.

sergei-maertens (approved):
nitpicking, suggested changes are just thinking out loud and I don't have strong feelings about them :)

src/lib/renderer/renderer.tsx
svenvandescheur: I can see how this could be confusing, and have not strong feelings either. Let's pull this in.

On a sidenote: do we have some kind of convention regarding restructuring? I feel it's a little context dependent to where it adds value.

sergei-maertens: in other places we pretty much always destructure (but only what's used), especially in the component itself, so rather prefer `const Foo = ({bar, baz}) => {...}` than `const Foo = (props) => {...}`)

PR #26, opened 2023-04-20
:hammer: Update build tooling

sergei-maertens:
* Replaced unmaintained microbundle-crl with pure tsc compilation
* Updated ESLint dependencies/rules
* Removed unused react-scripts/jsdom tests (we'll use storybook test runner instead)
* Fixed some eslint violations
* Removed the example boilerplate
* Removed react-scripts


.eslintrc
```diff
@@ -29,6 +29,7 @@
     "react/jsx-handler-names": 0,
     "react/jsx-fragments": 0,
     "react/no-unused-prop-types": 0,
-    "import/export": 0
+    "import/export": 0,
+    "no-use-before-define": 0
```
sergei-maertens, line 33: was getting errors about `React` not being defined, but that's not needed since the compiler inserts those already by itself depending on the `jsx` mode.

package.json
```diff
@@ -1,37 +1,50 @@
...
+    },
+    "./esm/": "./dist/esm/"
+  },
   "scripts": {
-    "build": "microbundle-crl --no-compress --format modern,cjs",
-    "start": "microbundle-crl watch --no-compress --format modern,cjs",
-    "test": "run-s test:unit test:lint test:build",
-    "test:build": "run-s build",
-    "test:lint": "eslint .",
-    "test:unit": "cross-env CI=1 react-scripts test --env=jsdom",
```
sergei-maertens, line 20: I removed this because there are no tests as of yet and we plan to rely on storybook test runner with the interaction tests anyway.

package.json
```diff
@@ -1,37 +1,50 @@
...
+    "url": "https://github.com/open-formulieren/formio-renderer.git"
   },
+  "author": "Maykin Media",
+  "license": "EUPL-1.2",
+  "bugs": {
+    "url": "https://github.com/open-formulieren/formio-renderer/issues"
+  },
+  "homepage": "https://github.com/open-formulieren/formio-renderer#readme",
   "peerDependencies": {
+    "formiojs": "~4.13.0 || ~4.14.0",
```
sergei-maertens, line 43: this should work for the formio utilities aspect without forcing the entire library as a CJS module in this library. Hopefully this gives us enough flexibililty to sort out the tree shaking.

src/components/textfield/textfield.component.tsx
```diff
@@ -48,7 +48,7 @@ export const TextField = (componentProps: ITextFieldProps): React.ReactElement =
   const _onChange = (event: React.FormEvent<HTMLInputElement>) => {
     setCharCountState(event.currentTarget.value.length);
     setPristineState(false);
-    onChange?.call(event.target, event);
+    onChange && onChange(event);
```
sergei-maertens, line 51: tsc was complaining about this, haven't properly validated if this still works though.

tsconfig.test.json
sergei-maertens, line 1: this did not appear to be used anywhere

package.json
```diff
@@ -1,37 +1,50 @@
...
-  },
-  "main": "dist/index.js",
-  "module": "dist/index.modern.js",
+  "main": "./dist/cjs/index.js",
+  "module": "./dist/esm/index.js",
+  "types": "./dist/esm/index.d.ts",
   "source": "src/index.tsx",
+  "exports": {
+    ".": {
+      "import": "./dist/esm/index.js",
```
svenvandescheur, line 11: Not opposed to this but I've had some issues in the past with packages defining esm modules for the import format (basically all modern es/js) which worries me a bit. Especially since node_modules libraries aren't always processed by loaders in webpack etc.

I'm also a bit worried if this is going to work at BD because of certain old systems (but I guess they just hit a build artifact).

src/lib/validation/validate.ts
svenvandescheur, line 1: Is this part of the build toolchain change?

src/lib/validation/validate.ts
```diff
@@ -74,6 +74,12 @@ export const getFormErrors = async (
   }
 };
 
+function isScalarValue(obj: IValues | Value | Values): obj is Value {
```
svenvandescheur, line 77: Isn't this just a primitive (more common name) check? Can't we simplify this too: `const isPrimitive = (value) => value !== Object(value)`?

svenvandescheur (changes requested):

.eslintrc
sergei-maertens: checked that, but it doesn't work and crashes out, the rule may have been deleted

it's also documented as discouraged to use: https://palantir.github.io/tslint/rules/no-use-before-declare/

It applies pretty much only to `var` keywords, which we typically don't use, so I'll leave this as is.

package.json
sergei-maertens: I made the setup the same like design-token-editor which is used with webpack without issues so far. The config should also be that doing `import {Foo} from '@open-formulieren/formio-renderer'` would default to CJS. The JS itself is transformed into es6.

Let me know if we run into any issues, for now both are supplied with minimal effort and I expect no problems.

src/lib/validation/validate.ts
sergei-maertens: yes, the compilar reported typing errors

src/lib/validation/validate.ts
sergei-maertens: the point of this function is that it provides type narrowing without casting it be a particular type, otherwise there are type errors due to the arbitrary JSON structure of the submission data.

svenvandescheur: I see the purpose of the function here but I still have concerns about the confusing naming (primitive should be name here) and its implementation. The "temporary" nature doesn't not change that aspect. Also: being a todo I think it should be either commented or an issue should be created for it.

sergei-maertens: There is actually a meaningful difference - scalar is the opposite from compounded (like arrays, objects...) and is intended to refer to actual, concrete values. From a mathematical point of view this is more correct, since primitives in JS also include symbols, which is most definitely not supported as a value.

I've created #27 to keep track of this.

svenvandescheur, line 77: Is there a reason why this is a regular function and not an arrow function adhering to the surrounding coding style?

src/lib/validation/validate.ts
```diff
@@ -74,6 +74,12 @@ export const getFormErrors = async (
...
+  if (Array.isArray(obj)) return false;
```
svenvandescheur, line 78: This seems redundant as typeof [] is object?

src/lib/validation/validate.ts
```diff
@@ -74,6 +74,12 @@ export const getFormErrors = async (
...
+  if (obj == null) return true;
```
svenvandescheur, line 79: Can this be a strict comparison?

src/lib/validation/validate.ts
svenvandescheur: This pull requests shouldn't have been merged without approval. Independently of the semantic discussion whether we should adhere to math or js concepts. I have some concerns regarding the implementation which should have been resolved through the proper process.

sergei-maertens: Yeah, had to look up exactly how to do type predicates with arrow functions and when I figured it out I found it completely unreadable.

See also https://stackoverflow.com/a/47847949

sergei-maertens: I merged it because it's fixing the  (reportedly) broken build chain, which was blocking any other feature and preventing us from using the formio utils which are known to have a correct implementation and our own implementation was undeniably insufficient

I shall adress the remaining comments tomorrow, either through justification as a comment or a follow up PR to correct them.

sergei-maertens: Renamed the function to `isSingleValue` because that's essentially what the type predicate expresses too. As added in the comments, this is still not 100% correct but that requires tweaking the Value types/interfaces extensively. I'm really thinking we should be able to map the static formio component types to the matching value types and use some typescript generics magic get specific types instead of having to deal with mismatching types (e.g. boolean for textfield does not make any sense).

sergei-maertens: Rewrote it as an arrow function in #29

As an aside, there are other cases where correctly typing React components with arrow functions and generics is not possible, where using a `function` declaration` is the only option. I also kind of fail the see why every function should be defined as an arrow function when plain functions work equally well and have benefits from being hoisted and being easier syntax to read, especially in combination with the syntax.

src/lib/validation/validate.ts
sergei-maertens: Yes, I had the logic the other way around first and forgot to clean this up.

There is some nuance where we might have to detect whether a value is an array or not, but that should rely on `component.multiple` rather than deriving it from untrusted value input.

src/lib/validation/validate.ts
sergei-maertens: Yep, I was thrown off by the convention that `foo == null` catches both `null` and `undefined`, but in this instance that's not relevant since `typeof undefined` is `undefined` so we only need to check for null explicitly. Addressed in #29

PR #33, opened 2023-05-01
Issue/#32 test formio compatibility

### LINKED ISSUE #32: Test compatibility with Formio

svenvandescheur:
closes #32


sergei-maertens (changes requested):
I do not think this approach with the vanilla formio renderer for the tests/stories is the correct one. It relies on developers writing the interaction tests for both flavours and keeping them in sync, and it's easy to forgot them. Instead I expected more something along the lines of a story factory function which you give the interaction/play function, and that returns both the react-based and vanilla formio renderer based stories automatically, e.g. something along the lines of:

```js
const [ReactRendererStory, VanillayRendererStory] = makeStory({
  template: <templateFunc>,
  play: async ({canvasElement}) => {
    const canvas = within(canvasElement);
    ...
  },
});

export ReactRendererStory;
export VanillayRendererStory;
```
If needed, you can then still override args/parameters per story.

This has a couple of advantages:

1. It guarantees that the same interaction tests pass for both the upstream and our own implementation
2. It guarantees that we don't forget to validate against the reference implementation
3. No/reduced duplication of story args/parameters/template functions
4. You can use a generic `renderWithVanillaFormio` function/utility that sorts out all the boilerplate in a single source of truth.

svenvandescheur:
> I do not think this approach with the vanilla formio renderer for the tests/stories is the correct one. It relies on developers writing the interaction tests for both flavours and keeping them in sync, and it's easy to forgot them. Instead I expected more something along the lines of a story factory function which you give the interaction/play function, and that returns both the react-based and vanilla formio renderer based stories automatically, e.g. something along the lines of:
> 
> ```js
> const [ReactRendererStory, VanillayRendererStory] = makeStory({
>   template: <templateFunc>,
>   play: async ({canvasElement}) => {
>     const canvas = within(canvasElement);
>     ...
>   },
> });
> 
> export ReactRendererStory;
> export VanillayRendererStory;
> ```
> 
> If needed, you can then still override args/parameters per story.
> 
> This has a couple of advantages:
> 
> 1. It guarantees that the same interaction tests pass for both the upstream and our own implementation
> 2. It guarantees that we don't forget to validate against the reference implementation
> 3. No/reduced duplication of story args/parameters/template functions
> 4. You can use a generic `renderWithVanillaFormio` function/utility that sorts out all the boilerplate in a single source of truth.

Yeah, I was already working on this I can probably use the [Form.io](http://form.io/) play functions as reference for both. will refactor.

package.json
```diff
@@ -45,6 +45,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.21.0",
+    "@formio/react": "^5.2.4-rc.1",
```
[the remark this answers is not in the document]
svenvandescheur: I can adjust the package but it worries me a bit, I based on the repository linked in on both packages link to the same repository but react-formio’s last update has been 2 years ago. I think it has seen a rename.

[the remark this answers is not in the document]
sergei-maertens: it was indeed renamed and it has plenty of issues, but this is the dependency used in both the backend and SDK so that is what you need to align with.

> Do we want to update the sdk to @react/formio?

No, we will update the SDK to use this library you're developing. We will definitely not invest resources in something that is going to be replaced anyway.

src/lib/format/placeholder.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Substitutes {{ placeholder }} with context[placeholder} in template string.
+ * TODO: This might need refactoring to suit translations.
```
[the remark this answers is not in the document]
sergei-maertens: Use what I suggested, as that is also what is used in the SDK. Remember that everything we do is only relevant in the context of Open Formulieren. We only care about the vanilla Formio behaviour to that extent.

Formio supports proving your own translation mapping and as you can see in the SDK that is what we do: https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/FormStep/index.js#L794

src/components/charcount/charcount.component.tsx
```diff
@@ -4,8 +4,8 @@ import React from 'react';
 
 export interface ICharCountProps extends IComponentProps {
   count: number;
-
   pristine: boolean;
+  showCharCount: boolean;
```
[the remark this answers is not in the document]
svenvandescheur: A number input can have showCharCount right? My idea was that input that can't have a certain reusable feature (showCharCount for checkbox) will simply not implement <CharCount />. Once <CharCount /> is called all the logic can be kept in one place utilizing the consistent (as possible) interface. This prevents this kind of logic from leaking into each implementation.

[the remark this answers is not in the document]
sergei-maertens: I don't see it in vanilla formio as a builder option: https://formio.github.io/formio.js/app/builder.html

And I'm 100% certain that we don't expose that option in our own builder configuration.

> This prevents this kind of logic from leaking into each implementation.

I do get this, but some (seemingly) re-usable components are actually very situational and different implementations may even deliberately choose to (not) support them, which is a valid choice. I can perfectly imagine that a webcomponent textfield or textarea implementation already has a similar feature implemented themselves.

At this time I'm really leaning towards letting the formio component type components (textfield, textarea, number...) do their own composition instead of us doing the "generic" composition. Because if we do it, then other implementations have no way to *not* use our wrappers except for overriding the wrapper component itself, which introduces a whole bunch of `if component.type` checks that they may not want to deal with.

For example, textfield, textarea, select etc. all likely need the outer `div` wrapper, but a checkbox component has very different markup where this outer wrapper div is not relevant and should not be present.

IMO the primary configuration interface should be the mapping of formio component type to the component to use. While being able to override the `label` / `description` etc. components, I do consider those a secondary layer that's further removed from public API and closer to private-ish API with more risks (i.e.: know what you're doing and be ready to deal with unpleasant changes between versions).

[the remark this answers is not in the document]
svenvandescheur: I'm not pointing to third-party (web) components here, but rather to our own components which will all benefit from a clearer separation. I agree that if we choose not to support certain features for subtypes (e.g.) number that those edge cases should be implemented in the components itself (e.g <TextField/>. But IMO the generic behavior should be implemented where it is used, the <CharCount/> itself.

I see an overlap with errors which isolates the display logic (whether the error tree should be rendered at all) as much as possible only adding a minimal additional interface (pristine) which should be derived from the components/form state.

By isolating such decision making a component can be much cleaner and risk of inconsistent behavior can be reduced.

Third party components have to take care of their own implementation. I don't think we need to expose partials to the outside world.

src/components/formio/columns/columns.component.tsx
```diff
@@ -1,11 +1,13 @@
-import {Component} from '@components';
 import {IComponentProps} from '@types';
 import clsx from 'clsx';
 import {ComponentSchema} from 'formiojs';
 import React from 'react';
 
+import {Component} from '../../utils';
```
sergei-maertens, line 6:
```suggestion
import {Component} from '@components/utils';
```
I find the relative paths so incredibly hard to reason about where exactly the thing is being imported from :grimacing:

src/components/formio/columns/columns.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+  <RenderComponent {...args} />
+);
+columns.args = {
+  component: {
+    key: undefined,
+    type: 'columns',
+    columns: [
+      {
+        size: 9,
+        components: [{type: 'content', html: 'Left column.'} as IContentComponent],
```
sergei-maertens, line 27: Looking at https://open-formulieren.github.io/open-forms-sdk/?path=/story/form-io-components-content--default-story&globals=locale:nl;locales.nl:Nederlands;locales.en:English and knowing how the form builder works (it always asks for a `key` property), I'd add `label` and `key` here too, otherwise you risk people copy pasting incomplete content component definitions.

Is the cast here really necessary btw, shouldn't this be possible with just the object definition without any casts?

src/components/formio/columns/columns.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+        size: 3,
+        components: [{type: 'content', html: 'Right column.'} as IContentComponent],
+      },
+    ],
+  },
+};
+columns.play = async ({canvasElement}) => {
+  const canvas = within(canvasElement);
+  expect(canvas.getByText('Left column.')).toBeTruthy();
+  expect(canvas.getByText('Right column.')).toBeTruthy();
```
sergei-maertens, line 39: I'd prefer `canvas.findBytext(...)` and `expect(...).toBeVisible()` here, avoiding mistakes/bugs that would result in the content being hidden by CSS/styles but still being present in the DOM.

src/components/formio/columns/columns.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+};
+columns.decorators = [
+  Story => (
+    <Formik initialValues={{columns: null}} onSubmit={() => {}}>
+      {Story()}
+    </Formik>
+  ),
+];
```
sergei-maertens, line 47: See also: https://github.com/open-formulieren/open-forms-sdk/blob/main/src/story-utils/decorators.js#L34

Definitely worthwhile to implement this as a re-usable decorator.

src/components/formio/columns/columns.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+  },
+};
+columns.play = async ({canvasElement}) => {
+  const canvas = within(canvasElement);
+  expect(canvas.getByText('Left column.')).toBeTruthy();
+  expect(canvas.getByText('Right column.')).toBeTruthy();
+};
+columns.decorators = [
+  Story => (
+    <Formik initialValues={{columns: null}} onSubmit={() => {}}>
```
sergei-maertens, line 43: the `initialValues` should just be `{}` here, as there are no actual data components. Columns itself (the formio component type) is only a layout container and will never be present in `submission.data`

src/components/formio/textfield/textfield.component.tsx
```diff
@@ -1,39 +1,38 @@
-import {CharCount, Component, Description, Errors, Label} from '@components';
 import {IComponentProps} from '@types';
 import clsx from 'clsx';
 import {ComponentSchema} from 'formiojs';
 import React, {useState} from 'react';
 
+import {CharCount, Component, Description, Errors, Label} from '../../utils';
```
sergei-maertens, line 6:
```suggestion
import {CharCount, Component, Description, Errors, Label} from '@components/utils';
```

src/components/formio/textfield/textfield.component.tsx
```diff
@@ -1,39 +1,38 @@
...
+
 export interface ITextFieldComponent extends ComponentSchema {
```
sergei-maertens, line 8: can you maybe add a `// TODO: replace with @open-formulieren/open-forms-types TextFieldSchema type when its available`?

src/components/formio/textfield/textfield.component.tsx
```diff
@@ -1,39 +1,38 @@
...
   id: string;
   inputMask: string;
+  key: string;
+  label: string;
   mask: string;
+  showCharCount: true;
```
sergei-maertens, line 14:
```suggestion
  showCharCount?: boolean;
```
not a literal, nor is it guaranteed to be present unfortunately (luckily enough undefined/null is falsy)

src/components/formio/textfield/textfield.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+const meta: Meta<typeof TextField> = {
+  title: 'Components / Formio / Textfield',
+  component: TextField,
+  decorators: [],
+  parameters: {},
+};
+export default meta;
+
+export const textfield: ComponentStory<
+  React.FC<IRenderComponentProps<ITextFieldComponent>>
```
sergei-maertens, line 18: not sure if it's possible, but can't you use some `typeof ...` stuff here to avoid re-declaring those types? That would make it a bit more self-correcting and more DRY.

src/components/formio/textfield/textfield.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+    label: 'first name',
+    mask: '',
+    showCharCount: true,
+    type: 'textfield',
+  },
+};
+textfield.play = async ({canvasElement}) => {
+  const canvas = within(canvasElement);
+  const input = canvas.getByLabelText('first name');
+  expect(canvas.getByText('Enter your first name')).toBeTruthy();
```
sergei-maertens, line 35: `.toBeVisible()` as an assertion expresses more intent & accessibility concerns

src/components/formio/textfield/textfield.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+  expect(canvas.queryByText('0 characters')).toBeNull();
+  await userEvent.type(input, 'The quick brown fox jumps over the lazy dog.', {delay: 10});
+  expect(canvas.queryByText('44 characters')).toBeTruthy();
+};
+textfield.decorators = [
+  Story => (
+    <Formik initialValues={{firstName: null}} onSubmit={() => {}}>
```
sergei-maertens, line 42:
```suggestion
    <Formik initialValues={{firstName: ''}} onSubmit={() => {}}>
```

src/components/utils/charcount/charcount.component.tsx
```diff
@@ -0,0 +1,14 @@
...
+export interface ICharCountProps {
+  value: string;
+}
+
+/**
+ * Reusable component showing `count` as number of characters typed into the component.
+ */
+export const CharCount: React.FC<ICharCountProps> = ({value}) => {
+  const className = clsx(`of-charcount`);
+  return <span className={className}>{value.length} characters</span>;
```
sergei-maertens, line 13: one next step after this is merged is to properly localize this with react-intl, because now the pluralization is not handled properly

src/components/utils/errors/errors.component.tsx
```diff
@@ -0,0 +1,30 @@
...
+export const Errors: React.FC<IErrorsProps> = ({componentId, errors}) => {
+  const className = clsx(`of-error-list`);
+  const listItemClassName = clsx(`of-error-list__item`);
+  const labelClassName = clsx(`of-error-list__message`);
+
+  return (
+    <ul className={className} aria-describedby={componentId}>
+      {errors?.map((error: string) => {
+        return (
+          <li key={error} className={listItemClassName}>
```
sergei-maertens, line 21: the errors list is not guaranteed to be a unique list of errors, so your key is not guaranteed to be unique. I'd incorporate the array index here.

src/components/utils/label/label.component.tsx
```diff
@@ -0,0 +1,20 @@
...
+}
+
+/**
+ * Label component.
+ */
+export const Label: React.FC<ILabelProps> = ({componentId, label}) => {
+  const className = clsx();
+
+  return (
+    <label className={className} htmlFor={componentId}>
```
sergei-maertens, line 16: note that if you plan to re-use this label for radio/checkboxes inputs the `htmlFor` will/should point to nothing

src/tests/utils/compatibility-stories-factory.tsx
```diff
@@ -0,0 +1,24 @@
...
+import {Args, ComponentStory} from '@storybook/react';
+import React from 'react';
+// @ts-ignore
+import {Form} from 'react-formio';
+
+/**
+ * Generates a tuple of 2 stories, one for Form.io and one For RenderForm. Both stories receive
+ * `args` and `play` function.
+ */
+export const compatibilityStoriesFactory = (args: Args, play: PlayFunction) => {
```
sergei-maertens, line 12: Let's make this function signature take an object with keys `args` and `play`, so that we mimick Storybooks interface. The advantage is that those names are very explicit then in the places where the factory is called.

sergei-maertens (changes requested):

src/components/utils/errors/errors.component.tsx
svenvandescheur: Ss there an actual case where show the same error twice? Given that resolving an error might render a new error "field foo is required" with index 0 becomes "field foo does not match pattern bar" with index 0 which might lead to unwanted behavior?

src/components/utils/charcount/charcount.component.tsx
svenvandescheur: Correct

src/components/utils/label/label.component.tsx
svenvandescheur: Correct, not implementing radios yet though.

src/tests/utils/compatibility-stories-factory.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const FormioStory = (args: Args) => <Form {...args} />;
+  FormioStory.args = args;
+  FormioStory.play = play;
+
+  const RenderFormStory: ComponentStory<typeof RenderForm> = args => <RenderForm {...args} />;
+  RenderFormStory.args = {
+    ...args,
+  };
```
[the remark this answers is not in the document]
svenvandescheur: Nah this is a typo, noticed it already when resolving the dependency conflict. I don't think we need (deep) cloning here. IMO for this case this can be a plain reference.

src/tests/utils/compatibility-stories-factory.tsx
svenvandescheur: I don't like this approach, it makes implementations less concise. I prefer the current shape.

src/tests/utils/compatibility-stories-factory.tsx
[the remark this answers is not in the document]
svenvandescheur: Actually on second though: I think the multiple branch might introduce some changes that might benefit from a clone.

src/components/utils/errors/errors.component.tsx
sergei-maertens: there shouldn't be but you can't rule it out. React still goes through the diffing algorithm if any of the props change, so if the error list itself is re-ordered, the `children` prop changes because the actual error text is different.

React itself advises to fall back to index if you cannot have a guaranteed stable identity/key for a given item in the iterable (while warning about the possible issues this has with mutating!).

Doing ```key={`${index}-${error}`}``` here would be guaranteed to be unique (though the index itself + child already does that)

src/tests/utils/compatibility-stories-factory.tsx
sergei-maertens: Leave it as is then, but if we add support for `parameters` and `decorators` too I'm not looking forward to updating the existing stories.

src/components/formio/columns/columns.component.tsx
svenvandescheur: Not opposed to introducing a separate namespace for this but lets address this elsewhere.

src/components/formio/textfield/textfield.component.tsx
svenvandescheur: weird, dunno why it's true here

svenvandescheur:
> `FORMIO_EXAMPLE` contains `select` components, which the renderer doesn't support yet. The interaction tests pass, because they only make assertions on the text fields. Which is fine, I guess.
> 
> But should the compatibility test check whether the produced element trees by both renderers are the same, or do we accept css breakage?

We don’t implement styling yet as it’s final implementation is yet to be decided.

CharString:
> We don’t implement styling yet as it’s final implementation is yet to be decided.

Sorry, I meant to say just that the elementtree structure differs between the two renders.

svenvandescheur:
> > We don’t implement styling yet as it’s final implementation is yet to be decided.
> 
> Sorry, I meant to say just that the elementtree structure differs between the two renders.

Correct the exact implementation is still open, door might be open for something like NLDS, and I think (can't find the discussion) a decision was made not to follow the exact structure/classname conventions put take the oportunity to refactor some stuff free of form.io limitations.

I'm not bound the the current implementation is the actual components since they need to be properly implemented with an actual design in mind. We probably wan't to track this separately.

sergei-maertens:
CSS/styling or element tree do not need to be equal, but the functional/interaction behaviour should be the same.

Clicking the label -> focus the field, dates -> using the same date format

PR #42, opened 2023-05-10
Feature/#31 ci

svenvandescheur:


sergei-maertens (changes requested):
See also #44 - this doesn't look polished yet to me.

sergei-maertens (approved):

Issue #49, opened 2024-03-11
Upgrade all the dependencies


nikkiysendoorn1:
Upgrade all the dependencies in the project.

PR #51, opened 2024-03-12
:arrow_up: [#49] update dependencies

nikkiysendoorn1:


.storybook/main.ts
sergei-maertens, line 1: Probably the approach from the builder works better here? https://github.com/open-formulieren/formio-builder/blob/main/.storybook/main.ts

.storybook/preview.ts
sergei-maertens, line 1: https://github.com/open-formulieren/formio-builder/blob/main/.storybook/preview.ts also has some additional type annotations that can/should be used here

package.json
```diff
@@ -17,15 +17,15 @@
   "scripts": {
     "build:esm": "tsc -p tsconfig.prod.json && tsc-alias -p tsconfig.prod.json",
     "build:cjs": "tsc -p tsconfig.prod.json --module commonjs --outDir dist/cjs && tsc-alias -p tsconfig.prod.json --outDir dist/cjs",
-    "build:storybook": "build-storybook --webpack-stats-json",
```
sergei-maertens, line 20: the `--webpack-stats-json` was used for some reason :thinking:

package.json
```diff
@@ -42,49 +42,49 @@
   "homepage": "https://github.com/open-formulieren/formio-renderer#readme",
   "peerDependencies": {
     "formiojs": "~4.13.0 || ~4.14.0",
-    "react": "^17.0.2"
+    "react": "^18.2.0"
   },
   "devDependencies": {
```
sergei-maertens, line 47: feel free to make the dev tools setup the same as the formio builder. I'd rather see ESLint added to both repositories at the same time instead of this mixed set up now.

This repo was started from a very outdated boilerplate, while the formio-builder one was crafted by hand. You can be bold ;)

sergei-maertens (changes requested):

.github/workflows/ci.yml
```diff
@@ -30,12 +30,12 @@ jobs:
...
           node-version-file: '.nvmrc'
-          cache: yarn
+          cache: npm
 
       - name: Build Storybook docs
         run: |
-          yarn install
-          yarn build-storybook --quiet
+          npm install
+          npm run storybook --quiet
```
sergei-maertens, line 38: ah, this is why it hangs :-) this is starting the SB dev server locally, but it should make a static build of storybook instead!

package.json
```diff
@@ -41,52 +45,40 @@
...
-    "eslint-plugin-react": "^7.17.0",
-    "eslint-plugin-standard": "^4.0.1",
-    "formiojs": "~4.13.0",
+    "@babel/core": "^7.24.0",
+    "@formatjs/ts-transformer": "^3.12.0",
+    "@storybook/addon-actions": "^8.0.0",
+    "@storybook/addon-essentials": "^8.0.0",
+    "@storybook/addon-interactions": "^8.0.0",
+    "@storybook/addon-links": "^8.0.0",
+    "@storybook/client-api": "^7.6.17",
```
sergei-maertens, line 57: is this dependency still needed? It's a lower major version than the main SB install (v8) so I expect problems.

You can run `npm why @storybook/client-api` to get more information :)

sergei-maertens (approved):
Thank you!

PR #91, opened 2025-03-21
[#73] Add content component

### LINKED ISSUE #73: Implement component type: content
**Reference**

https://open-formulieren.github.io/open-forms-sdk/?path=/story/form-io-components-vanilla-content--content

**Expected properties/features**

- [x] `html` (from WYSIWYG editor)
- [x] `customClass`
- [x] Check if we can run the content through DOMPurify to strip out potential XSS?

vaszig:
Closes #73


vaszig:
Tried to use the right css rules as far as I could. There are a lot hardcoded rules in `microscope-sass/lib/typography` library concerning `wysiwyg` and `body` which I do not know if we want to dive into and change all of these to NLDS at this PR. @sergei-maertens what are your thoughts?

package.json
```diff
@@ -66,6 +66,7 @@
     "@vitejs/plugin-react": "^4.3.4",
     "ajv": "^8.17.1",
     "babel-plugin-formatjs": "^10.5.34",
+    "dompurify": "^3.2.4",
```
sergei-maertens, line 69: I think we need to list this one in the `dependencies` rather than `peerDependencies` since it's a hard requirement.

src/registry/content/index.stories.tsx
```diff
@@ -0,0 +1,54 @@
+import {ContentComponentSchema} from '@open-formulieren/types';
```
sergei-maertens, line 1:
```suggestion
import type {ContentComponentSchema} from '@open-formulieren/types';
```

src/registry/content/index.tsx
```diff
@@ -0,0 +1,26 @@
+import {ContentComponentSchema} from '@open-formulieren/types';
+import {HTMLContent} from '@utrecht/component-library-react';
```
sergei-maertens, line 2: ooooh nice, didn't know this existed!

src/registry/content/Content.scss
```diff
@@ -0,0 +1,76 @@
+@use '@/scss/bem';
+@use '@utrecht/components/link';
+@use 'microscope-sass/lib/typography';
```
sergei-maertens, line 3: I'd expect a `@use '@utrecht/components/html-content;` rule too to load the CSS for that component, which enables us to specify and use design tokens for that.

src/registry/content/Content.scss
```diff
@@ -0,0 +1,76 @@
...
+ * Extensions of the utrecht-link component with OF-specific styles.
+ */
+.utrecht-link {
+  @include bem.modifier('openforms') {
+    font-family: var(
+      --of-utrecht-link-font-family,
+      var(--of-typography-sans-serif-font-family, var(--utrecht-document-font-family))
+    );
+    margin: 0;
+  }
```
sergei-maertens, line 27: I don't think the `font-family` rule is necessary - now would probably be a good time to use the standard NL DS design tokens.

Once we switch to the new renderer, I'm fairly certain we'll bump the SDK to a new major version (4.0) anyway so we can document the breaking changes.

src/registry/content/Content.scss
```diff
@@ -0,0 +1,76 @@
...
+
+  // swap the hover/no-hover text decoration rules compared to the non-modified variant
+  @include bem.modifier('openforms-hover') {
+    text-decoration: var(--utrecht-link-hover-text-decoration);
+
+    &:hover {
+      text-decoration: var(--utrecht-link-text-decoration);
+    }
+  }
```
sergei-maertens, line 36: this may not be necessary anymore, it was a workaround for CSS that had the wrong order and I think I reported that back upstream.

src/registry/content/Content.scss
sergei-maertens, line 3: `microscope-sass` is on the blocklist - it is not to be used in new code anymore (and the formio-renderer is new code). We consider it technical debt, because it relies on global variables which conflicts with the sass module system (`@use` vs. `@import`) and it hardcodes the font color/family etc. rather than allowing us to use design tokens.

src/registry/content/Content.scss
```diff
@@ -0,0 +1,76 @@
...
+    &:hover {
+      color: inherit;
+    }
+  }
+}
+
+.openforms-formio-content {
+  @include typography.wysiwyg;
+  @include typography.body;
+  @include typography.body--big;
```
sergei-maertens, line 50: this should be replaced with design token values for the utrecht components, or custom CSS rules that use our own (new?) design tokens instead of the hardcoded situation we now end up in.

sergei-maertens (changes requested):

src/registry/content/Content.scss
vaszig: I have added this but there are no design tokens for this. It only provides styles for the `utrecht-html` class (which is used in the `HTMLContent` component).

src/registry/content/Content.scss
vaszig: We already have design tokens for headings 1,2,3 and paragraph so I guess this is ok to be removed.

vaszig:
Have processed the feedback but I think we should discuss a bit some things concerning the design tokens when you have some time.

sergei-maertens (approved):

PR #95, opened 2025-04-13
Add low-level support for tooltips

### LINKED ISSUE #93: Implement tooltips (base)
Set up base components so that components can render their tooltips

* These must be accessible
* They must support our existing design tokens
* HTML must be escaped/sanitized

sergei-maertens:
Closes #93 

This sets up the low-level components to support tooltips next to labels.

This approach should be properly accessible to screenreaders and support keyboard navigation.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,6 +17,7 @@
       },
       "devDependencies": {
         "@babel/core": "^7.24.0",
+        "@floating-ui/react": "^0.27.7",
         "@formatjs/cli": "^6.5.1",
         "@fortawesome/fontawesome-free": "^6.7.2",
         "@open-formulieren/design-tokens": "^0.59.0",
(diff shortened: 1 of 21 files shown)


src/components/forms/InputGroup/InputGroup.scss
```diff
@@ -1,3 +1,8 @@
+// NOTE: the import order matters here, otherwise paragraph styles overwrite the styles
+// of form-fieldset
+@use '@utrecht/components/paragraph';
+@use '@utrecht/components/form-fieldset';
```
sergei-maertens, line 4: and yet it breaks :(

robinmolen (approved):
It look really nice! I tried it with my screen-reader (the default macOS VoiceOver) and it works really well. Also the color contrast is great!

One small improvement to the storybook tests:
- Maybe it would be nice to have a storybook entry with both the tooltip and error message, and that we check that the `describedBy` features both of the ids.

sergei-maertens:
For future reference - I stumbled on `expect(...).toHaveAccessibleErrorMessage(...)` and via that route learned about `aria-invalid` + `aria-errormessage` to refer to the ID of the error message, but NL DS did their research there already and concluded there's not sufficient support in screen readers.

Link from @robinmolen: https://a11ysupport.io/tech/aria/aria-errormessage_attribute

PR #96, opened 2025-04-14
Update label markup & styles in Storybook

sergei-maertens:
Following the example on https://nl-design-system.github.io/utrecht/storybook/\?path\=/docs/css_css-form-field--docs, we no longer render the form field label in an utrecht-paragraph, but instead we use a plain div with the utrecht-form-field__label class name.

This fixes the CSS precedence issues for the margin below the label, which both the paragraph and the form field try to set and the CSS rules that come last win.

These changes should be visually identical.

### DIFF
--- a/src/components/forms/EditGrid/EditGridItem.stories.tsx
+++ b/src/components/forms/EditGrid/EditGridItem.stories.tsx
@@ -1,6 +1,7 @@
 import type {Meta, StoryObj} from '@storybook/react';
 import {expect, fn, userEvent, within} from '@storybook/test';
 import {Paragraph} from '@utrecht/component-library-react';
+import '@utrecht/components/paragraph';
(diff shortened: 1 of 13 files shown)


src/components/forms/EditGrid/EditGridItem.stories.tsx
```diff
@@ -1,6 +1,7 @@
 import type {Meta, StoryObj} from '@storybook/react';
 import {expect, fn, userEvent, within} from '@storybook/test';
 import {Paragraph} from '@utrecht/component-library-react';
+import '@utrecht/components/paragraph';
```
sergei-maertens, line 4: added to ensure that the CSS is loaded for this story, since the Paragraph component is only used here and not in the component itself.

Issue #98, opened 2025-04-14
Implement component type: fieldset


sergei-maertens:
TODO (check if there's anything missing, there's already a component now)

sergei-maertens:
Checked, everything seems to be there

Issue #36, opened 2023-05-08
Implement repeating group (editgrid)


joeribekker:
See also: open-formulieren/open-forms#2441

**Tasks**

- [x] Implement React primitives/components
- [x] Find solution for icon library
- [x] Get MVP implementation that works for appointments flow & Formio component
- [x] Support validation errors & zod validation schema

**Expected properties/features for the formio component**

- [x] Nested components
- [x] `label`
- [x] `key`
- [x] `groupLabel`
- [x] `description`
- [ ] `tooltip` (postponed)
- [x] `hideLabel`
- [x] `disableAddingRemovingRows`
- [x] `addAnother` (text)
- [x] `saveRow` (text)
- [x] `removeRow` (text)

**Tougher challenges**

- [ ] Evaluate (conditional) visibility before rendering a nested component for each item
    - both the ValueDisplay
    - and the form field display
- [ ] Evaluating logic for an item inside the list of items
- [ ] Handle item-level validation errors (rather than errors about a field inside the item)

PR #112, opened 2025-04-17
Implement tooltip support for formio components

sergei-maertens:
Part of open-formulieren/open-forms#5133, #98 and #36

Extended the registry components that support tooltips and added stories for them.

### DIFF
--- a/src/registry/fieldset/index.tsx
+++ b/src/registry/fieldset/index.tsx
@@ -3,6 +3,7 @@ import clsx from 'clsx';
 
 import FormFieldContainer from '@/components/FormFieldContainer';
 import type {FormioComponentProps} from '@/components/FormioComponent';
+import Tooltip from '@/components/forms/Tooltip';
 import type {RegistryEntry} from '@/registry/types';
 
 import './Fieldset.scss';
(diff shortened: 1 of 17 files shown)


src/registry/fieldset/index.tsx
```diff
@@ -16,13 +17,21 @@ export interface FieldsetProps {
 }
 
 export const Fieldset: React.FC<FieldsetProps> = ({
-  // TODO: display tooltip, if specified
-  componentDefinition: {components, hideHeader = false, label},
+  componentDefinition: {components, hideHeader = false, label, tooltip},
   renderNested: FormioComponent,
 }) => {
   return (
     <fieldset className={clsx('openforms-fieldset', {'openforms-fieldset--no-header': hideHeader})}>
```
robinmolen, line 24: A bit unrelated to the tooltip, but...

1. I notice the class "openforms-fieldset--no-header" is only applied if `hideHeader` is set to true. However, if no `label` is set, then the header is also hidden. So should the condition for the class not be changed to `hideHeader && !!label`?
2. The `hideHeader` gives a "Deprecated" warning (set in the open-formulieren/types packages). So, is `hideHeader` correct? Or should the types project be updated?

robinmolen (approved):
2 comments about the fieldset, unrelated to the tooltip changes 😄.

Tooltip changes look good 👍

src/registry/fieldset/index.tsx
sergei-maertens: The label is required and enforced by the builder. Empty labels are not possible, unless someone messes with the JSON and then the renderer "breaking" is their well -deserved punishment 😉 

The deprecation has a reason in the types - I think there's a built in construct in Dormio that we missed at the time that we need to clean up at some point, but it's not critical. Or it's just that it should go in our own extensions namespace. Can be ignored for now!

robinmolen: 1. If label is required, and should always be a non-empty string, then `{!hideHeader && label && (...` can just be simplified to `{!hideHeader && (...` right?
2. Check..

sergei-maertens: 1. TS only guarantees that label is a string, not that it's not an empty string :grimacing: 

I'm a little bit concerned that we may be rendering empty `legend` nodes if someone manages to sneak in an empty string for the label (maybe via bad translations?) because that's making a bad case (someone messed with label JSON) even worse (screenreaders now get an empty legend in their tree), but yeah, this kinda contradicts my earlier point of "messing with the JSON and things breaking is then your own fault".

I'll simplify it.

sergei-maertens: Resolved via 4da3f218f5c44fc6359fa47606cb9c3ca4fd072c

PR #113, opened 2025-04-22
Implements the columns component

### LINKED ISSUE #72: Implement component type: columns
**Reference**

https://open-formulieren.github.io/open-forms-sdk/?path=/story/form-io-components-vanilla-columns--default

**Expected properties/features**

- [x] `columns`
- [x] `size`
- [x] `sizeMobile`
- [x] Nested components inside a `column`

sergei-maertens:
Closes #72

* Ported the CSS from the SDK and polished it
* Takes into account the validation schema, default values and visibility of nested components

### DIFF
--- a/src/registry/columns/index.tsx
+++ b/src/registry/columns/index.tsx
@@ -29,24 +29,20 @@ export const Columns: React.FC<ColumnsProps> = ({
             [`openforms-columns__column--span-mobile-${sizeMobile}`]: sizeMobile,
           })}
         >
-          Col {index + 1}
+          {components.map(nestedDefinition => (
+            <FormioComponent key={nestedDefinition.id} componentDefinition={nestedDefinition} />
+          ))}
         </FormFieldContainer>
       ))}
-
-      {/*<FormFieldContainer>
-        {components.map(nestedDefinition => (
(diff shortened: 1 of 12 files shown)


src/scss/responsive.scss
```diff
@@ -0,0 +1,10 @@
+// use microscope-sass library for the time being, but extract the responsive lib at some point.
+@forward 'microscope-sass/lib/responsive' with (
```
vaszig, line 2: So, is this done because it will be used used in the whole repo?

src/registry/columns/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+          {components.map(nestedDefinition => (
+            <FormioComponent key={nestedDefinition.id} componentDefinition={nestedDefinition} />
+          ))}
+        </FormFieldContainer>
+      ))}
+    </div>
+  );
+};
+
+const FieldsetComponent: RegistryEntry<ColumnsComponentSchema> = {
```
vaszig, line 40: This should be `Columns` right?

sergei-maertens: yep, copy paste error!

vaszig (approved):

PR #114, opened 2025-05-16
Implement (nested) simple conditionals

### LINKED ISSUE #59: Support simple conditional logic
Formio supports simple conditional logic in the form of:

```json
{
  "conditional": {
    "show": true,
    "when": "someOtherComponent",
    "eq": "someValue"
  }
}
```

which changes the visibility of a component depending on the value of the other component it references.

Notable challenges to take into account:

- [x] Ensure that this works with/within layout components like `fieldset` and `columns`
- [x] Implement the `clearOnHide` behaviour for components that get hidden. Take into account components that are hidden because their parent is hidden.
- [x] Ensure the behaviour in `editgrid` is consistent with the Formio reference (there are some odd mechanisms to refer to components in/outside of the editgrid)
- [x] Cycle detection
- [x] Component-specific value access/comparison (selectboxes, multiple: true)
- [ ] Ensure validation (schema) doesn't trip over hidden components

sergei-maertens:
Closes #59 partly

Added simple conditionals to editgrids, taking into account the correct evaluation scope.

### DIFF
--- a/src/reference/hidden.stories.ts
+++ b/src/reference/hidden.stories.ts
@@ -186,3 +186,102 @@ const {custom: ConditionallyNestedVisible, reference: ConditionallyNestedVisible
   });
 
 export {ConditionallyNestedVisible, ConditionallyNestedVisibleReference};
+
+const {custom: ConditionallyVisibleInEditGrid, reference: ConditionallyVisibleInEditGridReference} =
+  storyFactory({
+    args: {
+      components: [
+        {
+          type: 'textfield',
+          id: 'externalTrigger',
+          key: 'externalTrigger',
+          label: 'Trigger outside editgrid',
+        },
+        {
+          type: 'editgrid',
+          id: 'editgrid',
+          key: 'editgrid',
+          label: 'Edit grid',
+          disableAddingRemovingRows: false,
+          groupLabel: 'Item',
+          hideLabel: false,
+          // @ts-expect-error
+          inlineEdit: false,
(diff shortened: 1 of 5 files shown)


sergei-maertens:
@robinmolen I hope you can "make cheese of it" :grimacing: this was doing my head in

robinmolen (approved):
Okay.. yeah this broke my brain for a bit.. But after checking the usage of `setIn`, it's freaking nice!

I'm missing a proper documentation for this functionality (maybe a `registry/editgrid/index.mdx`?). Because when anyone has to touch this, they will bash their head to a wall 😅
And in this documentation, it would be nice to have some code examples of how the data changes (just to make it super clear that `editgrid: [object1, object2]` gets changed to `editgrid: object1` to do the editgrid item specific conditional logic)

sergei-maertens:
> Okay.. yeah this broke my brain for a bit.. But after checking the usage of `setIn`, it's freaking nice!
> 
> I'm missing a proper documentation for this functionality (maybe a `registry/editgrid/index.mdx`?). Because when anyone has to touch this, they will bash their head to a wall 😅 And in this documentation, it would be nice to have some code examples of how the data changes (just to make it super clear that `editgrid: [object1, object2]` gets changed to `editgrid: object1` to do the editgrid item specific conditional logic)

Done! I've added some sample component definitions and their matching submission data shapes + description how this shape changes in the scope of a single item.

PR #123, opened 2025-07-16
Implement textarea component

### LINKED ISSUE #109: Implement component type: textarea
TODO

robinmolen:
Closes #109

Implementation of the textarea component.

### DIFF
--- a/src/components/forms/Textarea/Textarea.stories.ts
+++ b/src/components/forms/Textarea/Textarea.stories.ts
@@ -101,7 +101,7 @@ export const NoAsterisks: Story = {
   name: 'No asterisk for required',
   decorators: [withFormSettingsProvider],
   parameters: {
-    renderSettings: {
+    formSettings: {
       requiredFieldsWithAsterisk: false,
     },
   },


src/components/forms/Textarea/Textarea.stories.ts
```diff
@@ -0,0 +1,188 @@
...
+    // Expect that the textarea grew to fit all content
+    expect(parseFloat(getComputedStyle(input).height)).toBeGreaterThan(initialHeight);
+
+    await userEvent.clear(input);
+    await userEvent.type(input, 'foo');
+    expect(input).toHaveFocus();
+    input.blur();
+
+    // Wait for the debounce
+    await sleep(500);
```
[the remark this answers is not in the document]
robinmolen: After discussing with Sergei, I've removed the debounce. With that, the `sleep()` is no-longer needed (and is removed)

src/components/forms/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,158 @@
...
+  isDisabled = false,
+  autoExpand = false,
+  placeholder,
+  tooltip,
+  ...extraProps
+}) => {
+  const {validateField} = useFormikContext();
+  const [{value, ...props}, {error = '', touched}] = useField<string | undefined>({
+    name,
+    type: 'text',
```
sergei-maertens, line 69: the `type` key only applies to `input` elements and Formik enables some special value casting behaviour based on its value: https://github.com/jaredpalmer/formik/blob/0e0cf9ea09ec864dd63c52cf775f862795ef2cf4/packages/formik/src/Formik.tsx#L926

here I think it's confusing to specify this key at all, since if you specfiy it, it should be something for `textarea`, but afaik textarea doesn't have a `type` attribute.

src/components/forms/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+      const newHeight = Math.max(minimalHeight, requiredHeight);
+      // Only update if new height is different from the current height
+      if (parseFloat(getComputedStyle(textarea).height) !== newHeight) {
+        textarea.style.height = `${newHeight}px`;
+      }
+    };
+
+    resizeTextarea();
+  }, [numberOfLineBreaks, autoExpand]);
```
sergei-maertens, line 116: shouldn't `textareaRef` also be here? it's used inside the hook.

src/components/forms/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+  return (
+    <FormField type="text" invalid={invalid} className="utrecht-form-field--openforms">
```
sergei-maertens, line 119:
```suggestion
    <FormField type="textarea" invalid={invalid} className="utrecht-form-field--openforms">
```
I think it controls the modifier

src/components/forms/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,158 @@
...
+          // ensure unsetting values doesn't ping-pong us between controlled/uncontrolled
+          // components
+          value={value ?? ''}
+          {...props}
+          onBlur={async e => {
+            props.onBlur(e);
+            await validateField(name);
+          }}
+          className={clsx('utrecht-textarea--openforms', {
+            'utrecht-textarea--openforms--no-resize': autoExpand,
```
sergei-maertens, line 140:
```suggestion
            'utrecht-textarea--openforms-no-resize': autoExpand,
```
probably best for namespace reasons

in BEM naming, the `__` and `--` separators can never repeat - you have a max of 3 levels deep (`block__element--modifier`)

src/components/forms/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,158 @@
...
+        parseFloat(borderBlockEnd)
+      );
+    };
+
+    const resizeTextarea = () => {
+      // Include the line without linebreak in the calculation
+      const numberOfLines = numberOfLineBreaks + 1;
+
+      const minimalHeight: number = calculateTextareaHeight(textarea.rows);
+      const requiredHeight: number = calculateTextareaHeight(numberOfLines);
```
sergei-maertens, line 106: can't you do something like `actualRows = Math.max(textarea.rows, numberOfLines);` and then run the height calculation only once?

I was also scanning for the `rows` prop, but it seems to be passed implicitly via the default props. I'd probably make this explicit and add it to the hook dependencies so you can use the prop directly instead of needing to grab it from the ref, it's the more react-ish way.

src/components/forms/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+      const newHeight = Math.max(minimalHeight, requiredHeight);
+      // Only update if new height is different from the current height
+      if (parseFloat(getComputedStyle(textarea).height) !== newHeight) {
+        textarea.style.height = `${newHeight}px`;
```
sergei-maertens, line 111: can we use `blockSize` here instead of `height`? I'm also concerned that using the `.style` API will cause CSP issues, unless this is also how Formio does it?

src/components/forms/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,158 @@
...
+      }
+    };
+
+    resizeTextarea();
```
sergei-maertens, line 115: I'm not sure why you define the function and then call it - can't you execute the function body in the hook immediately?

src/registry/registry.ts
```diff
@@ -10,6 +10,7 @@ import Fieldset from './fieldset';
 import PhoneNumber from './phoneNumber';
 import RadioField from './radio';
 import Selectboxes from './selectboxes';
+import textarea from './textarea';
```
sergei-maertens, line 13:
```suggestion
import TextArea from './textarea';
```
let's apply consistent casing

src/registry/textarea/index.stories.tsx
sergei-maertens, line 1: Can you add a story for `showCharCount: true`?

src/registry/textarea/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+import ValueDisplay from './ValueDisplay';
+import getInitialValues from './initialValues';
+import getValidationSchema from './validationSchema';
+
+export interface FormioTextareaProps {
+  componentDefinition: TextareaComponentSchema;
+}
+
+export const FormioTextarea: React.FC<FormioTextareaProps> = ({
+  componentDefinition: {key, label, description, tooltip, placeholder, validate},
```
sergei-maertens, line 15: `showCharCount` is not handled :)

sergei-maertens (changes requested):

src/components/forms/Textarea/Textarea.tsx
robinmolen: Huh... good point...

And yeah, good idea. I'll use the prop directly

src/registry/textarea/index.tsx
robinmolen: I'll add it to the component level (I think that would be better then the registry layer?). For the textfield I've added a separate ticket (https://github.com/open-formulieren/formio-renderer/issues/130)

src/registry/registry.ts
```diff
@@ -31,6 +32,7 @@ const REGISTRY: Registry = {
   checkbox: Checkbox,
   selectboxes: Selectboxes,
   radio: RadioField,
+  textarea: Textarea,
```
sergei-maertens, line 36: can you put this in the right order according to the UI?

Image

it goes before the checkbox. See also the same structure in the builder repo: https://github.com/open-formulieren/formio-builder/blob/6f3d309259bde6e686b20ea078a8220d22ff16bf/src/registry/index.ts#L68

src/components/forms/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,158 @@
...
+  ...extraProps
+}) => {
+  const {validateField} = useFormikContext();
+  const [{value, ...props}, {error = '', touched}] = useField<string | undefined>(name);
+  const id = useId();
+  const textareaRef = useRef<HTMLTextAreaElement | null>(null);
+
+  const invalid = touched && !!error;
+  const errorMessageId = invalid ? `${id}-error-message` : undefined;
+  const displayCharacterCount = showCharCount && maxLength !== undefined && value !== undefined;
```
sergei-maertens, line 84:
```suggestion
  const displayCharacterCount = showCharCount && value !== undefined;
```
you can enable `showCharCount` without setting a maximum length - in that case you just display the number of characters used. Only if both `showCharCount` and `maxLength` are provided do you show "x characters remaining".

src/components/forms/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,158 @@
...
+            defaultMessage="{remainingCharCount, plural,
+                            one {1 character remaining}
+                            other {{remainingCharCount} characters remaining}
+                          }"
+            values={{
+              remainingCharCount: maxLength - value.length,
+            }}
+          />
+        </span>
+      )}
```
sergei-maertens, line 149: Can you put this block in an `aria-live="polite"` container? I've attached https://github.com/open-formulieren/open-forms/issues/5460 as a sub-issue for this ticket.

sergei-maertens: And of course, update the messaging accordingly depending on whether there's a `maxLength` or not :)

robinmolen: haha, yeah i just saw the ticket. Will do

src/components/forms/Textarea/Textarea.tsx
robinmolen: That makes sense... 👍

src/components/forms/Textarea/Textarea.tsx
robinmolen: After some testing with the macOS screen reader, this will drive the users insane 😅 .

By just applying the `aria-live="polite"`, it will announce every update (which is also done for the textarea content), so you will be bombarded with the typed character and updated limit messages.

I think this is the right solution https://dav-idc.com/making-a-character-count-component-more-accessible/#coding-a-quieter-component.

sergei-maertens (approved):

PR #125, opened 2025-07-18
Add license plate component

### LINKED ISSUE #101: Implement component type: licenseplate
**expected properties**

- [x] label
- [x] description
- [x] tooltip
- [x] validate.required
- [x] validate.pattern
- [ ] multiple (later)

viktorvanwijk:
Closes #101

### DIFF
--- a/src/registry/licenseplate/index.tsx
+++ b/src/registry/licenseplate/index.tsx
@@ -1,0 +1,37 @@
+import type {LicensePlateComponentSchema} from '@open-formulieren/types';
+
+import TextField from '@/components/forms/TextField';
+import type {RegistryEntry} from '@/registry/types';
+
+import ValueDisplay from './ValueDisplay';
+import getInitialValues from './initialValues';
+import getValidationSchema from './validationSchema';
+
+export interface FormioLicensePlateProps {
+  componentDefinition: LicensePlateComponentSchema;
+}
+
+export const FormioLicensePlate: React.FC<FormioLicensePlateProps> = ({
+  componentDefinition: {key, label, description, tooltip, validate},
+}) => {
+  return (
+    <TextField
+      name={key}
+      label={label}
+      tooltip={tooltip}
+      description={description}
+      isRequired={validate.required}
+      pattern={validate.pattern}
+      placeholder="0-AAA-12"
+    />
+  );
+};
+
+const LicensePlateComponent: RegistryEntry<LicensePlateComponentSchema> = {
+  formField: FormioLicensePlate,
+  valueDisplay: ValueDisplay,
+  getInitialValues,
+  getValidationSchema,
+};
+
+export default LicensePlateComponent;
--- a/src/registry/licenseplate/ValueDisplay.scss
+++ b/src/registry/licenseplate/ValueDisplay.scss
@@ -1,0 +1,1 @@
+@use '@utrecht/components/ordered-list';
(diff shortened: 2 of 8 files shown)


src/registry/licensePlate/index.stories.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+import {FormioFormProps} from '@/components/FormioForm';
+import {renderComponentInForm} from '@/registry/storybook-helpers';
+import {withFormik} from '@/sb-decorators';
+
+import {FormioLicensePlate as LicensePlate} from './';
+import ValueDisplay from './ValueDisplay';
+
+export default {
+  title: 'Component registry / custom / licensePlate',
```
sergei-maertens, line 13:
```suggestion
  title: 'Component registry / custom / licenseplate',
```
(for parity with the `type` key value)

src/registry/licensePlate/index.tsx
sergei-maertens, line 1: can you rename the folder to `licenseplate` for consistent casing? the pattern is `src/registry/{component.type}` to make it easy to navigate the code

src/registry/licensePlate/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+}) => {
+  return (
+    <TextField
+      name={key}
+      label={label}
+      tooltip={tooltip}
+      description={description}
+      isRequired={validate.required}
+      pattern={validate.pattern}
+      placeholder=""
```
sergei-maertens, line 25:
```suggestion
      placeholder="XXX-XXX-XXX"
```
perhaps, to help with the pattern?

src/registry/licensePlate/validationSchema.ts
```diff
@@ -0,0 +1,27 @@
...
+const getValidationSchema: GetValidationSchema<LicensePlateComponentSchema> = (
+  componentDefinition,
+  intl
+) => {
+  const {key, validate} = componentDefinition;
+  const {required, pattern} = validate;
+
+  let schema: z.ZodFirstPartySchemaTypes = z
+    .string()
+    .regex(new RegExp(pattern), {message: intl.formatMessage(LICENSE_PLATE_INVALID_MESSAGE)});
```
sergei-maertens, line 21: I *think* we want to hardcode the pattern here instead of taking it from the component definition, which allows us to retro-actively validate license plates if the regex every changes in the future, without needing to update all existing form definitions in the backend. For type-safety and sanity checks, I'd probably do something like:

```ts
const LICENSE_PLATE_PATTERN: LicensePlateComponentSchema['validate']['pattern'] = '<regex here>';
const LICENSE_PLATE_REGEX = new RegExp(LICENSE_PLATE_PATTERN);
```
so that we have a module constant that only needs to initialize once.

sergei-maertens (changes requested):

src/registry/licenseplate/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+}) => {
+  return (
+    <TextField
+      name={key}
+      label={label}
+      tooltip={tooltip}
+      description={description}
+      isRequired={validate.required}
+      pattern={validate.pattern}
+      placeholder=""
```
sergei-maertens, line 25: Let's go with your placeholder suggestion :)

sergei-maertens (approved):

PR #126, opened 2025-07-18
Detect conditional logic cycles

### LINKED ISSUE #59: Support simple conditional logic

sergei-maertens:
Closes #59 (partially)

- [x] Implement cycle detection
    - [x] Simple components
    - [x] Layout components
    - [x] Edit grid
- [x] Wire up detection in top-level `FormioForm` component.

### DIFF
--- a/src/formio.ts
+++ b/src/formio.ts
@@ -99,11 +99,14 @@ function* iterDependencies(components: AnyComponentSchema[]): Generator<[string,
       }
 
       case 'editgrid': {
-        // TODO: we need to 'modify' the components here to take their scope into
-        // account -> this means making the full keys of items/components explicit.
-        // I don't think you can have cycles that span over scopes (outer scope -> item
-        // scope), but you can definitely have cycles within an item scope.
-        throw new Error('not implemented');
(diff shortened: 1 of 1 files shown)


robinmolen (approved):
Sadly, I couldn't come up with a components structure that would break your code 😉 

It looks real solid. Lets release it, and watch it in action!

PR #129, opened 2025-07-21
Conditional comparison complex types

sergei-maertens:
Partly closes #59 

TODO:

- [x] Implement comparison for selectboxes
- [x] Implement comparison for `multiple: true`
- [x] Pass component map to `isHidden` helper from `processVisibility` (put in context and cache?)
- [x] Defer comparison to specific component
- ~~Handle `number` component type references~~ Moved to #103

PR #134, opened 2025-07-25
Fix infinite render/state update in nested edit grids

### LINKED ISSUE #133: Nested editgrids can end up in an infinite render loop
Given a nested editgrid:

* the outer edit grid with key `outer` has a checkbox component and a nested `inner` edit grid
* the nested `inner` edit grid has three components:
    * a checkbox
    * a textfield, hidden when the `outer` checkbox is checked
    * a textfield, hidden when the `inner` checkbox is checked

Both textfields have `clearOnHide: true`

Now, when you expand all the edit grid items and toggle the check/uncheck the outer checkbox, you see two things:

* the initial value/state of the inner checkbox is reset
* in the debugger there are React warnings about maximum depth exceeded because of the state updates that lead to infinite render loops again

At no point was any of the "Save" buttons clicked. Clicking them both stops the infinite render.

sergei-maertens:
Closes #133

I could not reproduce this with a single edit grid, only with nested edit grids. Initially found via #129, but I managed to reproduce it based from `main` and create an isolated minimal test case.

**TODO**

- [x] Create regression test
- [x] Find root cause and fix it
- [x] Remove story once bug is fixed

### DIFF
--- a/src/registry/editgrid/visibility.ts
+++ b/src/registry/editgrid/visibility.ts
@@ -15,15 +15,15 @@ const applyVisibility: ApplyVisibility<EditGridComponentSchema> = (
   const initialValues = extractInitialValues(components, context.getRegistryEntry);
 
   // ensure we keep adding to the parent scope in case of nested edit grids
-  const outerEvaluationScope = context?.getEvaluationScope?.(values) ?? values;
+  const outerGetEvaluationScope = context?.getEvaluationScope ?? ((v: JSONObject): JSONObject => v);
 
   let items: JSONObject[] = getIn(values, key) ?? [];
   for (let index: number = 0; index < items.length; index++) {
     const itemValues = items[index];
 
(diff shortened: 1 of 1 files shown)


src/components/FormioForm.spec.tsx
```diff
@@ -440,3 +441,116 @@ describe('Updating form errors', () => {
...
+      expect(error!.message).toMatch(/Maximum update depth exceeded/);
+    }
+
+    expect(consoleErrorSpy.mock.calls).toHaveLength(0);
+    expect(error).toBeUndefined();
+
+    consoleErrorSpy.mockRestore();
+  });
+
+  test('Properly clears hidden fields in nested editgrids', async () => {
```
sergei-maertens, line 525: @robinmolen you were [right](https://github.com/open-formulieren/formio-renderer/pull/120#discussion_r2221894951) :sweat_smile:

robinmolen (approved):
Happy to help 😉
Looks good!

PR #137, opened 2025-08-01
Add number component

### LINKED ISSUE #103: Implement component type: number
Besides the standard component properties, it should support:

- [x] `suffix` with support for HTML `<sub>` and `<sup>`
- [x] `decimalLimit`
- [x] `allowNegative`
- [x] `validate.min`
- [x] `validate.max`
- [ ] `multiple`? Currently, this cannot be selected in the builder of this component.

**Other tasks**

- [x] Take into account the decimal separator based on the locale: `.` for English and `,` for Dutch.
- [x] Add test case to `hidden.stories.ts` for a conditional with `typeof eq === 'number'`

viktorvanwijk:
Closes #103

### DIFF
--- a/src/components/forms/NumberField/NumberField.stories.ts
+++ b/src/components/forms/NumberField/NumberField.stories.ts
@@ -1,0 +1,223 @@
+import type {Meta, StoryObj} from '@storybook/react';
+import {expect, userEvent, within} from 'storybook/test';
+
+import {withFormSettingsProvider, withFormik} from '@/sb-decorators';
+
+import NumberField from './NumberField';
+
+export default {
+  title: 'Internal API  / Forms / NumberField',
+  component: NumberField,
+  decorators: [withFormik],
+  parameters: {
+    formik: {
+      initialValues: {
+        number: null,
+      },
+    },
+  },
+} satisfies Meta<typeof NumberField>;
+
+type Story = StoryObj<typeof NumberField>;
+
+export const Default: Story = {
+  args: {
+    name: 'number',
+    label: 'Number',
+    description: 'This is a custom description for the number field',
+    isReadonly: false,
+    isRequired: false,
+  },
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+    const input = canvas.getByLabelText('Number');
+
+    // Assert that clicking on the label focuses the input
+    const label = canvas.getByText('Number');
+    await userEvent.click(label);
+    await expect(canvas.getByRole('textbox')).toHaveFocus();
+
+    // Assert that you are not able to type letters
+    await userEvent.type(input, 'foo');
+    await expect(input).toHaveDisplayValue('');
+  },
+};
+
+export const Readonly: Story = {
+  args: {
+    name: 'number',
+    label: 'Number',
+    description: 'This is a custom description for the number field',
+    isReadonly: true,
+  },
+  parameters: {
+    formik: {
+      initialValues: {
+        number: 10.5,
+      },
+    },
+  },
+};
+
+export const LocalisedWithDecimals: Story = {
+  args: {
+    name: 'number',
+    label: 'Number',
+    description: 'This is a decimal number formatted using the Dutch locale',
+    isReadonly: false,
+    isRequired: true,
+  },
+  parameters: {
+    formik: {
+      initialValues: {
+        number: 1.5,
+      },
+    },
+  },
+  globals: {
+    locale: 'nl',
+  },
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+    const input = canvas.getByLabelText('Number');
+
+    // Assert that you're able to provide a decimal number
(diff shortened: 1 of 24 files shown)


src/components/forms/NumberField/_input-container.scss
```diff
@@ -0,0 +1,77 @@
...
+// Note: this was ported from the SDK
+
+$affixes: 'prefix', 'suffix';
+
+@mixin override($property, $modifier) {
+  $default: var(--of-input-container-affix-#{$property});
+  #{$property}: var(--of-input-container-#{$modifier}-#{$property}, #{$default});
+}
+
+.#{'openforms-input-container'} {
```
sergei-maertens, line 12: this essentially makes up its own component `InputContainer` and the styling code + wrapper JSX probably belong in their own place in `src/components/forms`. They can be applied in a broader context than just number fields.

Other than that, this is a bit of an odd construct in SASS, easier to understand and read is:


```suggestion
.openforms-input-container {
```

src/components/forms/NumberField/NumberField.stories.ts
```diff
@@ -0,0 +1,231 @@
...
+  title: 'Internal API  / Forms / NumberField',
+  component: NumberField,
+  decorators: [withFormik],
+  parameters: {
+    docs: {
+      source: {
+        type: 'dynamic',
+        excludeDecorators: true,
+      },
+    },
```
sergei-maertens, line 18: I think this is no longer necessary

src/components/forms/NumberField/NumberField.tsx
```diff
@@ -0,0 +1,185 @@
...
+   * Required fields get additional markup/styling to indicate this validation requirement.
+   */
+  isRequired?: boolean;
+  /**
+   * Disabled fields get marked as such in an accessible manner. Note that Form.io uses
+   * 'disabled' in the component definition, but we would like to have the behaviour of
+   * a read-only field so we use the readOnly prop internally.
+   *
+   */
+  isDisabled?: boolean;
```
sergei-maertens, line 37: already here the prop should be renamed to `isReadonly` - at the Formio layer (registry) we can do: `isReadonly={component.disabled}` where the translate Formio's oopsie to proper semantics

src/registry/number/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+    key,
+    label,
+    description,
+    tooltip,
+    validate,
+    decimalLimit,
+    allowNegative,
+    prefix,
+    suffix,
+    disabled,
```
sergei-maertens, line 25:
```suggestion
    disabled: isReadonly,
```
to alias it :)

src/registry/number/ValueDisplay.tsx
```diff
@@ -0,0 +1,20 @@
+import type {NumberComponentSchema} from '@open-formulieren/types';
+import {useIntl} from 'react-intl';
+
+export interface ValueDisplayProps {
+  componentDefinition: NumberComponentSchema;
+  value: number | null;
+}
+
+const ValueDisplay: React.FC<ValueDisplayProps> = ({
+  // @ts-ignore
```
sergei-maertens, line 10: why is the ignore necessary?

src/registry/number/ValueDisplay.tsx
```diff
@@ -0,0 +1,20 @@
...
+  componentDefinition,
+  value,
+}) => {
+  const {locale} = useIntl();
+
+  if (value === null) return '';
+  return new Intl.NumberFormat(locale).format(value);
```
sergei-maertens, line 17: can't we use https://formatjs.github.io/docs/react-intl/components/#formattednumber? (it uses the `Intl` APIs under the hood)

src/registry/number/ValueDisplay.tsx
```diff
@@ -0,0 +1,20 @@
...
+import {useIntl} from 'react-intl';
+
+export interface ValueDisplayProps {
+  componentDefinition: NumberComponentSchema;
+  value: number | null;
+}
+
+const ValueDisplay: React.FC<ValueDisplayProps> = ({
+  // @ts-ignore
+  componentDefinition,
```
sergei-maertens, line 11: I'd think we can/should take the number of decimal places configuration into account for the number formatting here.

sergei-maertens (changes requested):
Nice one!

src/registry/number/ValueDisplay.tsx
viktorvanwijk: Good point :)

Fun fact: I added a `decimalLimit: 3` to the story and used a value with more decimals, and it already worked, so I was surprised and confused. Apparently 3 is also the default number of decimal places for the `maximumFractionDigits` prop of `Intl.NumberFormat` :upside_down_face:

sergei-maertens: ah the classic "my code works, but why?" :smile:

PR #149, opened 2025-08-26
Implement onChange prop for FormioForm

### LINKED ISSUE #79: Implement `onValuesChanged` prop/callback
The SDK/parent component must be able to react to value changes inside the form (because of user input) so that backend logic calls can be scheduled (see #78). Ideally, this would be via a prop `onValuesChanged={newValues => {...}}`.

I'm not sure yet at the moment if we only need to react to value changes because of user input (could be wired up through Formik's `handleChange`) or to all changes (in which case we'd probably need to run a `useEffect` hook).

**Tasks**

- [x] Figure out which events should trigger the callback
- [x] Add the `onValuesChanged` prop
- [x] Implement the callback behaviour

sergei-maertens:
Closes #79

Anytime the Formik values state changes (because of user input or a calculated/programmatic update), the onChange handler is invoked if/when it is provided.

* Went with the more idiomatic `onChange` prop
* Events: every update is reported back, there's no filtering on user input vs. automatic update (for simplicity)
* Looked into `startTransition` etc. to check if there was some kind of natural/built in debounce behaviour in case quick value updates arrive in succession, but there's doesn't seem to be anything. It's probably better for the SDK to debounce and/or use `useTransition` to make it clear that stuff is happening in the background.

### DIFF
--- a/src/0001-index.mdx
+++ b/src/0001-index.mdx
@@ -8,7 +8,7 @@ import {Meta} from '@storybook/addon-docs/blocks';
 
 ## Quickstart
 
-**NOTE** - this library is still in active development and _NOT_ stable.
+**NOTE** - this library is still in active development and public API _may_ change.
 
 **High level API**
 
@@ -41,6 +41,7 @@ const MyComponent: React.FC = () => (
   <IntlProvider messages={{}} locale="nl" defaultLocale="nl">
     <FormioForm
       components={FORMIO_FORM.components}
+      onChange={values => console.log(values)}
       onSubmit={values => console.log(values)}
       id="my-formio-form"
       requiredFieldsWithAsterisk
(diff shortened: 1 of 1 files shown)


src/components/FormioForm.spec.tsx
robinmolen, line 1: Should we add a test for components nested in editgrid? 🙂

src/components/FormioForm.mdx
```diff
@@ -17,6 +17,17 @@ The `onSubmit` callback receives the entered form values when the HTML form is s
 
 <Canvas of={FormioFormStories.Example} />
 
+## Subscribing to value changes
+
+The `onChange` prop receives _every_ value change in the form state, which covers:
```
robinmolen, line 22: (Editgrid, gotta love it 😉)

Editgrid would only trigger after saving the editgrid item (nested form, that doesn't register on the top level). Should we state that value changes in a editgrid item only register after editgrid item saving?

robinmolen (commented):
Looks nice! I really like the `FormValuesObserver` implementation.

I have 2 small questions/comments about, you guessed it, editgrid 😉

PR #151, opened 2025-08-28
Implement modal component

### LINKED ISSUE #169: Implement Modal component
The modal component is used by the partner and children form components.

For the implementation was decided to use a basic html/css setup, with https://github.com/modelbrouwers/modelbrouwers/blob/main/src/js/components/modals/Modal.tsx and https://github.com/modelbrouwers/modelbrouwers/blob/main/src/sass/components/_modal.scss as inspiration. See original comment: https://github.com/open-formulieren/formio-renderer/issues/119#issuecomment-3182720700.

robinmolen:
Closes #169 

The modal uses the native HTML dialog element with all its features.

To open the modal, we need a useEffect hook. The HTML dialog element does support a `open` property, but this can only be used to show the dialog inline without modal pop-up functionalities.

Opening the modal automatically sets the correct aria-tags.

The modal can be closed using the [esc] key, and close button.

### DIFF
--- a/src/components/modal/Modal.scss
+++ b/src/components/modal/Modal.scss
@@ -9,7 +9,7 @@ $modal-padding-block-end: var(--of-modal-padding-block-end, 40px);
 $modal-padding-inline-start: var(--of-modal-padding-inline-start, 40px);
 $modal-padding-inline-end: var(--of-modal-padding-inline-end, 40px);
 
-// Take halve the modal padding, and turn it negative
+// Take half the modal padding, and turn it negative
 $modal-close-icon-offset-block-start: calc($modal-padding-block-start * -0.5);
 $modal-close-icon-offset-inline-end: calc($modal-padding-inline-end * -0.5);
 
@@ -96,7 +96,7 @@ body:has(.openforms-modal[open]) {
     opacity: 0.5;
     font-size: 150%;
 
-    // This button is only used to show an icon. With this the icon is centralized.
+    // This button is only used to show an icon. With this the icon is centered.
     line-height: 1;
 
     &:focus {
--- a/src/components/modal/Modal.stories.tsx
+++ b/src/components/modal/Modal.stories.tsx
@@ -69,8 +69,8 @@ export const CloseModal: Story = {
 
       // The `closeModal` arg should have been called, and the modal should no-longer be
       // visible.
-      await expect(args.closeModal).toHaveBeenCalled();
-      waitFor(() => {
+      await waitFor(() => {
+        expect(args.closeModal).toHaveBeenCalled();
         expect(dialog).not.toBeVisible();
       });
     });
@@ -84,8 +84,10 @@ export const CloseModal: Story = {
 
       // The `closeModal` arg should have been called, and the modal should no-longer be
       // visible.
-      await expect(args.closeModal).toHaveBeenCalled();
-      expect(dialog).not.toBeVisible();
+      await waitFor(() => {
+        expect(args.closeModal).toHaveBeenCalled();
+        expect(dialog).not.toBeVisible();
+      });
     });
   },
 };


src/components/modal/Modal.scss
sergei-maertens, line 1: I'd expect that the appearance of the modal can be controlled with design tokens/CSS variables.

src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,80 @@
...
+import {useEffect, useId, useRef} from 'react';
+import {useIntl} from 'react-intl';
+
+import Icon from '@/components/icons';
+
+import './Modal.scss';
+
+// @TODO is this needed? Or does the HTMl dialog handle this natively?
+// Copied logic from the current SDK implementation.
+// Used to prevent scrolling in the page, when the modal is open.
```
sergei-maertens, line 11: Looks like you can do this in pure CSS: https://stackoverflow.com/a/77722504

src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,80 @@
...
+    if (!modal) {
+      return;
+    }
+
+    isOpen ? modal.showModal() : modal.close();
+  }, [isOpen]);
+
+  return (
+    <dialog
+      className="modal"
```
sergei-maertens, line 55:
```suggestion
      className="openforms-modal"
```
let's stick to our existing class name pattern and properly namespace things

sergei-maertens (changes requested):

src/components/modal/Modal.tsx
robinmolen: Ah ofcourse, simple `overflow: hidden` is the classic way of doing this... 🤦

src/components/modal/Modal.stories.tsx
```diff
@@ -0,0 +1,81 @@
+import type {Meta, StoryObj} from '@storybook/react-vite';
+import {useState} from 'react';
+import type {ArgsStoryFn, Renderer} from 'storybook/internal/csf';
```
sergei-maertens: which part of `internal` in this import is not clear?

Aren't you better of implementing this as a `import type {Decorator} from `@storybook/react-vite';`? 

You can then also grab the `useArgs` hook from the [story context](https://storybook.js.org/docs/writing-stories/decorators#context-for-mocking), instead of needing `useState` here, because changing the `isOpen` arg of the story will not have any effect when the modal open/closed state is tracked by `useState`.

src/components/modal/Modal.scss
```diff
@@ -0,0 +1,103 @@
+@use '@utrecht/components/button';
+@use '@/scss/bem';
+@use '@/scss/responsive';
```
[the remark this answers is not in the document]
robinmolen: I'm not sure why this comment is not marked as outdated, as i've changed the heading-2 for heading-1 (conform the SDK setup)..

Anyway, I'll separate the imports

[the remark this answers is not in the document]
sergei-maertens: I don't think it should be heading-1 - analytics people complain and get very nervous when they see more than one h1 in the page markup

src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,76 @@
...
+        </Heading2>
+      )}
+
+      <SubtleButton
+        className="openforms-modal__close"
+        type="button"
+        onClick={closeModal}
+        aria-label={closeButtonLabel}
+      >
+        <Icon icon="close" aria-hidden />
```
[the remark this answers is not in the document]
sergei-maertens: no - utrecht-icon is really meant to wrap around the icon (svg/font...) itself, see https://nl-design-system.github.io/utrecht/storybook/?path=/docs/css_css-icon--docs

Issue #104, opened 2025-04-14
Implement component type: postcode


sergei-maertens:
**Properties**

- [x] `description`
- [x] `label`
- [x] `tooltip`
- [x] `inputMask` -> as placeholder
- [ ] `multiple` (postponed)
- [x] `validate.pattern`
- [x] `validate.required`

PR #162, opened 2025-10-02
[#104] Add postcode component

SonnyBA:


SonnyBA:
These changes make use of the [PostcodeComponentSchema](https://github.com/open-formulieren/types/blob/73e4be07dc1a850a6a0d156e8bf6438ea177b7d2/src/formio/components/postcode.ts#L43) which is marked as deprecated. I'm not sure if we want to update this type or we want to remove it's usage here :thinking:.

sergei-maertens:
> These changes make use of the [PostcodeComponentSchema](https://github.com/open-formulieren/types/blob/73e4be07dc1a850a6a0d156e8bf6438ea177b7d2/src/formio/components/postcode.ts#L43) which is marked as deprecated. I'm not sure if we want to update this type or we want to remove it's usage here 🤔.

There are some limitations why we can't remove this yet :) Not sure what exactly they were, but the goal is indeed to eventually remove this component type and instead use some variation of textfield.

Okay to ignore the deprecation for now!

src/registry/postalcode/index.stories.tsx
```diff
@@ -0,0 +1,244 @@
...
+} satisfies Meta<typeof PostalCodeField>;
+
+type Story = StoryObj<typeof PostalCodeField>;
+
+export const MinimalConfiguration: Story = {
+  args: {
+    componentDefinition: {
+      id: 'component1',
+      type: 'postcode',
+      key: 'my.postalcodefield',
```
sergei-maertens, line 25: please use `postcode` instead of `postal code` - it *is* valid English :)

src/registry/postalcode/index.stories.tsx
```diff
@@ -0,0 +1,244 @@
...
+      initialValues: {
+        my: {
+          postalcode: '',
+        },
+      },
+    },
+  },
+};
+
+export const WithPlaceholder: Story = {
```
sergei-maertens, line 45: this is effectively the same as above, so this story can be dropped.

src/registry/postalcode/validationSchema.ts
```diff
@@ -0,0 +1,31 @@
...
+import type {GetValidationSchema} from '@/registry/types';
+
+const POSTAL_CODE_PATTERN: PostcodeComponentSchema['validate']['pattern'] =
+  '^[1-9][0-9]{3} ?(?!sa|sd|ss|SA|SD|SS)[a-zA-Z]{2}$';
+const POSTAL_CODE_REGEX = new RegExp(POSTAL_CODE_PATTERN);
+
+const POSTAL_CODE_INVALID_MESSAGE = defineMessage({
+  description: 'Validation error for postal code.',
+  defaultMessage: 'Invalid Dutch postal code',
+});
```
sergei-maertens, line 14: The Formio translations in the backend have the right message:


```suggestion
const POSTAL_CODE_INVALID_MESSAGE = defineMessage({
  description: 'Validation error for postal code.',
  defaultMessage: 'The submitted value does not match the postcode pattern: 1234 AB',
});
```
(Dutch translation: `De opgegeven waarde voldoet niet aan het postcode formaat: 1234 AB`, you can already include that in the `nl.json` file)

src/registry/registry.ts
```diff
@@ -14,6 +14,7 @@ import IBAN from './iban';
 import LicensePlate from './licenseplate';
 import Number from './number';
 import PhoneNumber from './phoneNumber';
+import PostalCodeField from './postalcode';
```
sergei-maertens, line 17:
```suggestion
import Postcode from './postalcode';
```

sergei-maertens (changes requested):
Can you change the directory name from `postalcode` to `postcode` so that it exactly matches the value of the `type` key?

PR #164, opened 2025-10-07
Mock current Date in some stories

sergei-maertens:
Chromatic otherwise detects new changes everytime it snapshots these stories again, because they use the real current Date. Mocking these provides stability for the snapshots.

This restores the changes from the previous build: https://www.chromatic.com/test?appId=67d017859163e5b55e120f90&id=68e518cee92fc829b3b7455f so the changed snapshots here/now are intentional.

I went for sinonjs/fake-timers since that's the API that vitest implements and it's also used by @storybook/test-runner.


viktorvanwijk (approved):
Ah yeah, makes sense :)

PR #165, opened 2025-10-08
[#110] Add time component

SonnyBA:


SonnyBA:
The current implementation allows the user to specify the second but I'm not sure if that is needed for this component.

src/registry/timefield/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  return (
+    <TimeField
+      name={key}
+      label={label}
+      tooltip={tooltip}
+      description={description}
+      isRequired={validate?.required}
+    />
+  );
+};
```
sergei-maertens: I think we can also do something with passing down the min/max times if they're in the validation configuration

src/registry/timefield/ValueDisplay.tsx
```diff
@@ -0,0 +1,60 @@
...
+    time = new Date();
+    time.setHours(parseInt(hours));
+    time.setMinutes(parseInt(minutes));
+    time.setSeconds(parseInt(seconds));
+  }
+
+  return (
+    <FormattedTime
+      value={time}
+      format="hh:mm:ss"
```
sergei-maertens: I played around a bit with the various options - I think the single prop `timeStyle="short"` instead of format/hour/minute/second would be sufficient and cover localization in multiple languages (it also omits the seconds).

However, it conflicts with https://onzetaal.nl/taalloket/tijden-noteren - the time is reported as `04:15` for example instead of `4.15 uur` which we'd prefer.

I'm not sure if we can do anything with the `format` string here, but I'm tempted to stick with `timeStyle="short"` for now and revisit later.

sergei-maertens (changes requested):
Can you rename the directory name from `timefield` to `time` so that it matches the `component.type` value?

PR #172, opened 2025-10-14
Fix radio inputs not being isolated between radio input items

### LINKED ISSUE #171: Radio components inside editgrids are not isolated from each other
Discovered/reported by Laurens in in https://github.com/open-formulieren/open-forms-sdk/issues/863

When you have a repeating group with a radio component, interacting with it in one item affects other items (when they are both in edit mode!).

Root cause: the `name` attribute is picked up and native browser radio group handling kicks in.

sergei-maertens:
Closes #171

Major rework needed to fight against the browser :(

**TODO**

- [x] Fix type information, there are some lies now

### DIFF
--- a/src/components/forms/EditGrid/EditGrid.tsx
+++ b/src/components/forms/EditGrid/EditGrid.tsx
@@ -1,13 +1,17 @@
 import {ButtonGroup} from '@utrecht/button-group-react';
 import {FormField, PrimaryActionButton} from '@utrecht/component-library-react';
-import {FieldArray, getIn, useFormikContext} from 'formik';
-import {useState} from 'react';
+import {FieldArray, getIn, setIn, useFormikContext} from 'formik';
+import type {FormikErrors} from 'formik';
+import {useMemo, useState} from 'react';
 import {FormattedMessage} from 'react-intl';
+import type {ValidationError} from 'zod-formik-adapter';
 
+import FieldConfigProvider from '@/components/FieldConfigProvider';
 import HelpText from '@/components/forms/HelpText';
 import Label from '@/components/forms/Label';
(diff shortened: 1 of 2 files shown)


src/components/forms/EditGrid/EditGrid.tsx
```diff
@@ -202,4 +214,124 @@ function EditGrid<T extends {[K in keyof T]: JSONValue} = JSONObject>({
...
+      const itemValues: T = getIn(values, namePrefix);
+
+      try {
+        await validateCallback(index, itemValues);
+      } catch (err: unknown) {
+        // only handle errors we know
+        if (!isValidationError(err)) throw err;
+        err.inner.forEach(problem => {
+          problem.path = `${namePrefix}.${problem.path}`;
+        });
```
robinmolen: Does this not need to be a Array.map?

robinmolen (approved):
Edit grid really is a proper pain..

I think this is correct, I haven't managed to break it 😄.

Maybe its a good idea to add storybook tests that the errors are still shown correctly in this situation (i've made a quick setup, and it seems fine, but still good to cover)

sergei-maertens:
> Edit grid really is a proper pain..
> 
> I think this is correct, I haven't managed to break it 😄.
> 
> Maybe its a good idea to add storybook tests that the errors are still shown correctly in this situation (i've made a quick setup, and it seems fine, but still good to cover)

I was going to do this, but checked and there already is a story with a validation error from the schema build through the formio component validation options in `?path=/story/component-registry-special-editgrid--with-validation-schema`

There are also stories for the low-level edit grid component that pass errors, and the important part in this rework was that the external API would not change (still pass in the same structure of data/errors etc.). I don't see which additional stories we can make in storybook tbh.

PR #175, opened 2025-10-17
chore: upgrade nl-design-system CSS packages

sergei-maertens:
Same as #174 but with a proper Chromatic build :)

Closes #174


sergei-maertens:
Nothing concerning in the Chromatic visual regression tests - and this fixes even a bug with the textarea border colors!

PR #180, opened 2025-10-22
Support async plugin validation

### LINKED ISSUE #76: Support asynchronous plugin validation
A number of components support `validate.plugins` which invokes a backend validation plugin.

There should be a mechanism (callback based) to call a validation plugin, and the SDK is then responsible for implementing the API call.

**Tasks**

- [ ] Expose `async runPluginValidation` hook through render context & add matching prop to `FormioForm`
- [ ] Pass the callback to the zod validation schema for supported components
- [ ] Document how to use it (from the SDK perspective)

sergei-maertens:
Closes #76

Implemented the async validation mechanism. The formio-renderer is agnostic about the actual HTTP API calls, the SDK is responsible for implementing the expected interface.

TODO

- [x] Check the interaction with #176 once it's merged

### DIFF
--- a/src/components/FormioForm.mdx
+++ b/src/components/FormioForm.mdx
@@ -155,7 +155,7 @@ const FormStep: React.FC = () => {
 ## Providing the `validatePluginCallback`
 
 Many of the Formio component types have been extended to support _plugin_-based backend validation -
-this is a mechanism were the backend implements the validation logic which can be called via an API
+this is a mechanism where the backend implements the validation logic which can be called via an API
 endpoint. Each plugin has an `identifier`, which is a just a plain string.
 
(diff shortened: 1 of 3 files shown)


sergei-maertens:
> Minor things and questions :)
> 
> Some components (select, radio, addressNL) have all the validation schema unit tests updated from `schema.safeParse` to `schema.safeParseAsync`, but others only have the newly added test use the async version. Any particular reason for this, or just didn't bother updating everything?

didn't bother - those code paths don't hit the async `superRefine` and are not affected. Under the hood the async function is called for the real validation in Formik, and that's also checked with some end-to-end tests.

sync code reads a tad easier and is simpler, but the primary reason is avoiding unnecessary diff noise

PR #183, opened 2025-10-24
:bug: Fix number and currency fields not showing validation errors

viktorvanwijk:
From https://github.com/open-formulieren/open-forms-sdk/issues/863

The `name` property was not passed to `NumericFormat` -> the field name was missing from the event in the `onBlur` handler -> `touched` was not set inside Formik's `onBlur` -> error was not shown.

I'm not entirely sure how `touched` is set in the stories, though, because the errors are shown there :thinking:

### DIFF
--- a/src/components/forms/NumberField/NumberField.tsx
+++ b/src/components/forms/NumberField/NumberField.tsx
@@ -129,7 +129,7 @@ const NumberField: React.FC<NumberFieldProps> = ({
 }) => {
   name = useFieldConfig(name);
   const {validateField} = useFormikContext();
-  // Note that we have a custom onValueChange handler, so `value` and `onBlur` is the only relevant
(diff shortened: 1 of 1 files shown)


sergei-maertens (approved):

viktorvanwijk:
> > I'm not entirely sure how touched is set in the stories, though, because the errors are shown there
> 
> https://github.com/open-formulieren/formio-renderer/blob/8058983860edcd87a7329ae301d14d51e52d0569/src/components/forms/NumberField/NumberField.stories.ts#L11
> and
> 
> https://github.com/open-formulieren/formio-renderer/blob/8058983860edcd87a7329ae301d14d51e52d0569/src/components/forms/NumberField/NumberField.stories.ts#L235
> 
> the storybook decorator picks up the `initialTouched` which helps for presentation in Storybook, but yeah, it's a bit fragile if the input is not actually touched :)

Thanks :)

Issue #102, opened 2025-04-14
Implement component type: map


sergei-maertens:
TODO

Mimic current SDK implementation:
- [x] Support placing point, line, and polygon data
- [x] Support search location, and placing point on result (and centralize and zoom on result)
- [x] Support WMS tile layers (extra shapes on top of map)
- [x] Support custom WMTS tile layer (map background)
- [x] Support lazy loading

PR #184, opened 2025-10-24
[#102] map component

SonnyBA:
Fixes: #102 

~TODO:~

- [x]     ~Location permissions~
- [x]     ~Location search~
- [x]     ~Lazy loading~
- [x]     ~Summary patch: https://github.com/open-formulieren/formio-renderer/pull/184#issuecomment-3460943138~
- [x]     ~Translations~
- [x]     ~Documentation~
- [x]     ~Add field to registry~

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,3 +6,4 @@ export * from '@/components/forms';
 export * from '@/components/Button';
 export {default as LoadingIndicator} from '@/components/LoadingIndicator';
 export {default as Modal} from '@/components/modal';
+export {default as LeafletMap} from '@/components/map/LeafletMap';


src/components/forms/MapField/index.ts
sergei-maertens: Do we really need to build/expose this as a low-level field? I think I'd rather to for `src/components/map` as a base and implement the low-level stuff there, and then combine it with the field-stuff in the registry.

There's *a lot* of code here that's directly related to maps, but only tangentially to form fields/Formik. I'd like to separate those more 🤔

SonnyBA:
The builds report a decrease in coverage, I don't think we can test more than is already tested though, as the missing coverage seems to be related to using the shapes with multiple points and functions that Leaflet calls:

- [Polyline behavior](https://app.codecov.io/gh/open-formulieren/formio-renderer/pull/184?src=pr&el=tree&filepath=src%2Fcomponents%2Fmap%2FdrawPolylineControl.ts&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=open-formulieren#diff-c3JjL2NvbXBvbmVudHMvbWFwL2RyYXdQb2x5bGluZUNvbnRyb2wudHM=),  these tests were also absent in the SDK and do not seem to be testable
- [Right clicking behavior](https://app.codecov.io/gh/open-formulieren/formio-renderer/pull/184?src=pr&el=tree&filepath=src%2Fcomponents%2Fmap%2FMapField.tsx&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=open-formulieren#diff-c3JjL2NvbXBvbmVudHMvbWFwL01hcEZpZWxkLnRzeA==), this was discussed earlier, these tests were also absent in the SDK and do not seem to be testable, see lines 293-298
- [Location control code](https://app.codecov.io/gh/open-formulieren/formio-renderer/pull/184?src=pr&el=tree&filepath=src%2Fcomponents%2Fmap%2FLeafletMapLocationControl.tsx&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=open-formulieren#diff-c3JjL2NvbXBvbmVudHMvbWFwL0xlYWZsZXRNYXBMb2NhdGlvbkNvbnRyb2wudHN4), these functions are defined so that Leaflet can call these

src/components/map/NearestAddress.tsx
```diff
@@ -0,0 +1,58 @@
...
+}
+
+/**
+ * Retrieve and display the nearest address label for the selected coordinates.
+ */
+const NearestAddress: React.FC<NearestAddressProps> = ({coordinates}) => {
+  // XXX: this would benefit from caching, but rather than rolling our own we should
+  // probably look into useQuery/tanstack react query
+  const formSettings = useFormSettings();
+  const mapNearestLookup = formSettings?.componentParameters?.map?.mapNearestLookup;
```
sergei-maertens: can you also prepare a (draft) PR in the SDK that provides/injects this? You can then already validate/test it via `npm run build && npm pack` + installing that tar.gz locally.

src/registry/map/map.mdx
```diff
@@ -0,0 +1,72 @@
...
+
+## Overlays
+
+To configure various overlays for the map, the `overlays` property can be specified. This allows for
+example, images and SVG's to be shown on the map over certain areas.
+
+```
+import 'MapComponent' from '@open-formulieren/formio-renderer';
+
+const overlayers = [
```
sergei-maertens:
```suggestion
const overlays = [
```

SonnyBA:
@sergei-maertens I'll use this open PR for the low-level component. Through 635315f I refactored the component and removed all the form related code and the Formik bindings.

We could move even more code to the registry component and make it similar to how the [`ValueDisplay`](https://github.com/open-formulieren/formio-renderer/pull/184/commits/ecfbe531ec571e431c70b9edcf3768ee0d1c3224#diff-a381581a1c5d3a3e9dc7ac32e52be000f421c305069677d0271156c833444e73L19) was implemented earlier (stripped down to a minimal map), although that will require us to pass through child elements from the registry component (for the location controls for example) to this component.

sergei-maertens (changes requested):
I think what's still missing here is in `src/index.ts` adding the export for `LeafletMap` so that the SDK can use that too and we can effectively replace most of the map code in the SDK!

PR #192, opened 2025-10-30
File (upload) component - presentational components

### LINKED ISSUE #99: Implement component type: file
**Features**

- [x] Handle single vs. multiple files
- [x] Client-side drag and drop uploads
- [x] Extract file metadata + validate eagerly (on drop?)
- [x] Upload hook for backend
- [x] TXT validation response disply -> later: also accept JSON errors
- [x] file size validation/checks + good display
- [x] keep track of temporary uploads
- [x] allow deleting the uploads
- ~~support camera functionality to take picture (check business logic from Formio)~~ looks like `webcam: false` disables this
- [x] Flexible layout control (Open Forms theme vs. Den Haag theme & help text positioning)
- [x] Make it a lazy component (`React.Lazy` for bundle splitting)

Incorporate as much design feedback as possible from DH (confirm final designs!):

Image

References:
* https://github.com/orgs/nl-design-system/discussions/347
* https://github.com/orgs/nl-design-system/discussions/381

sergei-maertens:
Partly closes #99 - define presentational components.

No user interaction and Formik field state updates are handled here yet, nor is the upload / drag and drop behaviour - that will come in a later PR.

This PR focuses on the markup, layout and styling for the upload component, roughly matching the designs provided by Den Haag. it is fully expected that we'll have to iterate on the styling, especially because the designs are not finalized yet.

The TODOs in the code can stay for now, but must be removed before a new release / closing the ticket, as there may be backwards compatibility challenges to still figure out.

Implements:

* Upload file preview, including 'pending' state for ongoing uploads
* List of uploaded files
* Rough layout with the label, drop zone, file upload list, help text, validation errors
* First pass for accessibility, but this will need some proper testing once everything is wired up.

### DIFF
--- a/i18n/compiled/en.json
+++ b/i18n/compiled/en.json
@@ -51,6 +51,12 @@
       "value": "index"
     }
   ],
+  "1AgCMs": [
+    {
+      "type": 0,
+      "value": "Remove"
+    }
+  ],
   "1TAtDi": [
     {
       "offset": 0,
@@ -153,6 +159,20 @@
     {
       "type": 0,
       "value": "MM"
+    }
+  ],
+  "8uM5Vx": [
+    {
+      "type": 0,
+      "value": "Remove '"
+    },
+    {
+      "type": 1,
+      "value": "fileName"
+    },
+    {
+      "type": 0,
+      "value": "'"
     }
   ],
   "9SgPC6": [
@@ -429,6 +449,12 @@
       "value": "Add another"
     }
   ],
+  "ZGuxfW": [
+    {
+      "type": 0,
+      "value": "File name"
+    }
+  ],
   "abKEBH": [
     {
       "type": 0,
@@ -489,6 +515,12 @@
       "value": "."
     }
   ],
+  "eAmrdi": [
+    {
+      "type": 0,
+      "value": "Loading..."
+    }
+  ],
   "lVNV/d": [
     {
       "type": 0,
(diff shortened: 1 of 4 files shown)


src/registry/file/UploadedFileList.scss
```diff
@@ -0,0 +1,14 @@
...
+  display: flex;
+  flex-direction: column;
+  row-gap: var(--of-openforms-uploaded-file-list-row-gap);
+
+  list-style: none;
+  padding: 0;
+  margin: 0;
+}
+
+// TODO: remove before merge!
```
vaszig: Same here

src/registry/file/UploadedFile.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+import * as UploadedFileStories from './UploadedFile.stories';
+
+<Meta of={UploadedFileStories} />
+
+# `UploadedFile`
+
+The `UploadedFile` component is used to display the files that have been uploaded by the end user.
+
+It can display the upload status and for succesful uploads provides a control to remove the file
```
vaszig: The same is happening when we have errors correct?

src/registry/file/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+            className="openforms-file-upload__dropzone"
+            aria-describedby={errorMessageId}
+            tabIndex={0}
+            role="button"
+          >
+            <Paragraph>FILE UPLOAD HERE</Paragraph>
+          </div>
+        )}
+
+        {!!value.length && (
```
vaszig: Generally should we use this instead of `value.length > 0`? just thinking that this is neater and more readable I guess from checking the length.

vaszig (commented):
Seems nice to me, don't have something to mention except for some thoughts/questions as always.

PR #195, opened 2025-11-03
Non-editable editgrid item in isolation mode

### LINKED ISSUE #122: Implement component type: children
Implementing Children component based on SDK setup.

Aiming for a non-modal setup. Exploring editgrid possibilities, for inline children editing

robinmolen:
Partly closes #122

Non-editable editgrid items in isolation mode cannot use `namePrefix` to retrieve their data from the formik context. This is because the `namePrefix` namespace, is the namespace of the isolated formik form state, which isn't available for non-editable editgrid items.

This PR solves this issue by setting the `namePrefix` to the globally available namespace, for non-editable editgrid items. This way, these editgrid items can still retrieve their value from the closest formik state.

For editgrid is this not-yet a problem, as all editgrids are hardcoded to be isolated and always editable. For possible future editgrid features this ensures that same functions of retrieving item values can be used for editable and non-editable items.

Now, for components that use the basic structure of editgrid, this creates possibilities to further control edit permissions. In the case of the children component, this allows the user to select an child (updating a `selected` property) while not being able to change the child data.

### DIFF
--- a/src/components/forms/EditGrid/EditGrid.stories.tsx
+++ b/src/components/forms/EditGrid/EditGrid.stories.tsx
@@ -185,7 +185,7 @@ export const NonEditableItemInIsolation: Story = {
     emptyItem: undefined,
     // The first item is uneditable, the second item is editable.
     canEditItem: (_, index) => index === 1,
-    validate: () => Promise.resolve(),
+    validate: async () => {},
   },
   parameters: {
     formik: {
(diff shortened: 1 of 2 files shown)


src/components/forms/EditGrid/EditGrid.stories.tsx
```diff
@@ -158,6 +160,49 @@ export const AddingItemInIsolationMode: Story = {
...
+export const NonEditableItemInIsolation: Story = {
+  args: {
+    enableIsolation: true,
+    getItemHeading: undefined,
+    getItemBody: () => <ItemBody />,
+    description: '',
+    emptyItem: undefined,
+    // The first item is uneditable, the second item is editable.
+    canEditItem: (_, index) => index === 1,
+    validate: () => Promise.resolve(),
```
sergei-maertens:
```suggestion
    validate: async () => {},
```
Use a consistent style of doing async/await - we want to hide the `Promise` concept as much as possible.

sergei-maertens (changes requested):

PR #196, opened 2025-11-03
Implement/port children component

### LINKED ISSUE #122: Implement component type: children

robinmolen:
Closes #122 

~Depending on #195~ Has been resolved

Implementing the children component inside the formio-renderer

### DIFF
--- a/src/registry/children/Children.scss
+++ b/src/registry/children/Children.scss
@@ -1,13 +1,8 @@
 @use '@utrecht/components/table';
 @use '@utrecht/components/custom-checkbox';
 
-@use '@/scss/bem';
-
 .openforms-children {
-  @include bem.modifier('with-manual-creating') {
-    .utrecht-form-field-description--openforms-errors {
-      // Add spacing between the error message and the "add child" button
-      margin-block-end: var(--of-utrecht-form-field-description-errors-margin-block-end, 8px);
-    }
+  .utrecht-form-field-description--openforms-errors + .utrecht-button {
+    margin-block-start: var(--of-children-error-button-spacing, 8px);
   }
 }
--- a/src/registry/children/index.tsx
+++ b/src/registry/children/index.tsx
@@ -1,5 +1,4 @@
 import type {ChildrenComponentSchema} from '@open-formulieren/types';
-import {clsx} from 'clsx';
 import type {FormikErrors} from 'formik';
 import {FieldArray, useFormikContext} from 'formik';
 import {useId, useState} from 'react';
@@ -70,9 +69,7 @@ export const FormioChildrenField: React.FC<FormioChildrenFieldProps> = ({
       isInvalid={invalid}
       hasTooltip={!!tooltip}
       aria-describedby={[descriptionId, errorMessageId].filter(Boolean).join(' ') || undefined}
-      className={clsx('openforms-children', {
-        'openforms-children--with-manual-creating': canAddChildrenManually,
-      })}
+      className="openforms-children"
     >
       <FieldArray name={key} validateOnChange={false}>
         {arrayHelpers => (


vaszig:
@robinmolen Before I dive into this I have a question (from a quick look in the UI tests)..

I guess it was discussed with Sergei and decided that we will use EditGrid for this?In the SDK we use a table for the children and this was the initial design (an editGrid was used only for the second step if it exists and this is decided by the backend/logic action).

robinmolen:
> @robinmolen Before I dive into this I have a question (from a quick look in the UI tests)..
> 
> I guess it was discussed with Sergei and decided that we will use EditGrid for this?In the SDK we use a table for the children and this was the initial design (an editGrid was used only for the second step if it exists and this is decided by the backend/logic action).

I mis interpreted a comment from Sergei about exploring EditGrid for this component. I'll rework this, to follow the SDK design and implementation

src/registry/children/ChildModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+import {BSNField, DateOfBirthField, FirstNamesField} from './subFields';
+import type {ExtendedChildDetails} from './types';
+import {buildChildSchema} from './validationSchema';
+
+const ChildModalSaveButton = () => {
+  const {submitForm} = useFormikContext<unknown>();
+  return (
+    <PrimaryActionButton
+      type="button"
```
sergei-maertens: with the reworked modal (rendering in a portal), you can use `type="submit"` and substantially simplify things -  `useFormikContext` is no longer necessary then

src/registry/children/index.stories.tsx
```diff
@@ -0,0 +1,507 @@
...
+    const canvas = within(canvasElement);
+
+    await step('Validate initial state', () => {
+      const checkboxes = canvas.getAllByRole('checkbox');
+      expect(checkboxes).toHaveLength(3);
+
+      // Only the selected child should be checked
+      expect(checkboxes[0]).not.toBeChecked();
+      expect(checkboxes[1]).not.toBeChecked();
+      expect(checkboxes[2]).toBeChecked();
```
sergei-maertens: can we give the checkboxes a proper accessible label (use the child first name?) and use those in the queries here?

src/registry/children/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+export interface FormioChildrenFieldProps {
+  componentDefinition: ChildrenComponentSchema;
+}
+
+export const FormioChildrenField: React.FC<FormioChildrenFieldProps> = ({
+  componentDefinition: {key, label, description, tooltip, enableSelection},
+}) => {
+  const id = useId();
+  const {getFieldProps, setFieldTouched, errors, touched} = useFormikContext();
+  const {value: children} = getFieldProps<ExtendedChildDetails[]>(key);
```
sergei-maertens: don't forget to use `@/hooks:useFieldConfig` on `key` to apply any prefixes that may be set from an edit grid.

src/registry/children/ChildrenTable.tsx
```diff
@@ -0,0 +1,167 @@
...
+                  aria-label={intl.formatMessage(
+                    {
+                      description: "Children component: child 'selected' aria label",
+                      defaultMessage: '{firstname} should be included in the form submission',
+                    },
+                    {
+                      firstname: child.firstNames,
+                    }
+                  )}
+                  checked={child.selected}
```
sergei-maertens: Okay good that there is a label, I don't think this is the optimal label text yet though.

```suggestion
                  aria-label={intl.formatMessage(
                    {
                      description: "Children component: child 'selected' aria label",
                      defaultMessage: 'Include {firstname}?',
                    },
                    {
                      firstname: child.firstNames,
                    }
                  )}
                  checked={child.selected}
```

I don't think the user needs to be bothered with "form submission" concepts.

src/registry/children/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const [childToEdit, setChildToEdit] = useState<ExtendedChildDetails | undefined>();
+
+  const error: FormikErrors<FormValues>[string] | string = getIn(errors, key);
+
+  // check if there's an error for the children field *itself* rather than one of the
+  // subkeys
+  const childrenError: string | undefined = typeof error === 'string' ? error : undefined;
+
+  const itemsTouched: ExtendedChildDetails[] | undefined = getIn(touched, key, undefined);
```
sergei-maertens: this type is wrong - `FormikTouched` is a (nested) object with field names as keys and `true | false` as value.

I'd also strongly recommend avoiding `getIn` unless there's no alternative, and instead use the formik helpers like `getFieldMeta` which give you proper derived types, e.g. in the file component:

```tsx
const {touched} = getFieldMeta<FormikFileUpload[]>(name);
```

Here you would really benefit from not tracking individual item touched states, but only track the touched state of the file component itself, similarly to the file component.

You can then explicitly set the field as a whole touched as soon as:

* a child is edited (`onSubmit` callback part for the modal)
* a child is added (`onSubmit` callback part for the modal)
* a child is removed (`onRemove` callback part, where you wire up the `arrayHelpers.remove`)
* checkbox blur for any child

src/registry/children/validationSchema.spec.ts
```diff
@@ -0,0 +1,232 @@
...
+    [{bsn: '111222333', firstNames: '', dateOfBirth: '1980-12-12'}],
+    [{bsn: '111222333', firstNames: 'Joe', dateOfBirth: ''}],
+
+    [{bsn: '', firstNames: 'Joe', dateOfBirth: ''}],
+    [{bsn: '', firstNames: 'Joe', dateOfBirth: '1980-12-12'}],
+
+    [{bsn: '', firstNames: '', dateOfBirth: '1980-12-12'}],
+  ])(
+    'does not accept children with missing BSN, first name or date of birth (value: %s)',
+    async (...value) => {
```
sergei-maertens: the spread syntax is tripping me up here. Can't this be:


```suggestion
  test.each([
    [{bsn: '', firstNames: '', dateOfBirth: ''}],

    [{bsn: '111222333', firstNames: '', dateOfBirth: ''}],
    [{bsn: '111222333', firstNames: '', dateOfBirth: '1980-12-12'}],
    [{bsn: '111222333', firstNames: 'Joe', dateOfBirth: ''}],

    [{bsn: '', firstNames: 'Joe', dateOfBirth: ''}],
    [{bsn: '', firstNames: 'Joe', dateOfBirth: '1980-12-12'}],

    [{bsn: '', firstNames: '', dateOfBirth: '1980-12-12'}],
  ])(
    'does not accept children with missing BSN, first name or date of birth (value: %s)',
    async (value: ChildDetails[]) => {
```

src/registry/children/index.stories.tsx
```diff
@@ -0,0 +1,507 @@
+import type {ChildrenComponentSchema} from '@open-formulieren/types';
+import type {Meta, StoryObj} from '@storybook/react-vite';
+import {expect, userEvent, within} from 'storybook/test';
+
+import ValueDisplay from '@/registry/children/ValueDisplay';
+import type {ExtendedChildDetails} from '@/registry/children/types';
```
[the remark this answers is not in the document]
robinmolen: Ah chips, missed some. Will do!

src/registry/children/ChildrenTable.tsx
```diff
@@ -0,0 +1,167 @@
...
+    <SubtleButton
+      onClick={removeChild}
+      hint="danger"
+      aria-label={intl.formatMessage(
+        {
+          description: "Children component: children table 'delete child' accessible label",
+          defaultMessage: 'Delete child with BSN: {bsn}',
+        },
+        {bsn: child.bsn}
+      )}
```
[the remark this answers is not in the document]
robinmolen: updated it

src/registry/children/index.tsx
robinmolen: Ah yeah, good one..

src/registry/children/ChildrenTable.tsx
robinmolen: Mhh, yeah thats better. Even just wording it as a question, makes it a lot better

src/registry/children/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+export const FormioChildrenField: React.FC<FormioChildrenFieldProps> = ({
+  componentDefinition: {key, label, description, tooltip, enableSelection},
+}) => {
+  const id = useId();
+  const {getFieldProps, setFieldTouched, errors, touched} = useFormikContext();
+  const {value: children} = getFieldProps<ExtendedChildDetails[]>(key);
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const [childToEdit, setChildToEdit] = useState<ExtendedChildDetails | undefined>();
+
+  const error: FormikErrors<FormValues>[string] | string = getIn(errors, key);
```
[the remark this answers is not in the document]
robinmolen: About the typing changes: i completely agree. I was having some difficulty myself with the typing of this thing, and used a definition of another component.

Either way, your setup for the files component works really nice. In addition with replacing the formik global errors with `getFieldMeta` errors, this has become a lot better.

I'm not sure about your final comment, about changing `ExtendedChildDetails` with `ChildDetails`. Wouldn't it be confusing if we use `ExtendedChildDetails` as type for the component values, but use `ChildDetails` for the errors? Also, the errors could in the future contain error messages for the data added in the `ExtendedChildDetails`, which will need to be covered.

src/registry/children/ChildrenTable.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+const EditChildModal: React.FC<EditChildModalProps> = ({child, onSubmit}) => {
+  const intl = useIntl();
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const closeModal = () => setIsModalOpen(false);
+
+  return (
+    <>
+      {/* Make sure to only render the modal when it's open,
+      so we don't have multiple modals in the DOM */}
```
sergei-maertens: what's the issue with multiple (hidden) modals? Each have their own identity and state. Is there an accessibility impact?

robinmolen: Okay, checking with VoiceOver this doesn't seems to be a problem. I'll fix it in the storybook test, and remove the comment

src/registry/children/BareCheckbox.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const [{value, ...props}] = useField<boolean | undefined>({
+    name,
+    type: 'checkbox',
+  });
+  return (
+    <UtrechtCheckbox
+      className="utrecht-form-field__input utrecht-custom-checkbox utrecht-custom-checkbox--html-input utrecht-custom-checkbox--openforms"
+      appearance="custom"
+      aria-label={ariaLabel}
+      // Without a default value, formik thinks the field behind the checkbox is actually
```
sergei-maertens: I need to understand better what this means, because in this form the checkbox component is an unmanaged component 🤔

src/registry/children/Children.scss
```diff
@@ -0,0 +1,13 @@
+@use '@utrecht/components/table';
+@use '@utrecht/components/custom-checkbox';
+
+@use '@/scss/bem';
+
+.openforms-children {
+  @include bem.modifier('with-manual-creating') {
```
sergei-maertens: do we need this modifier at all?

If there are no errors, the node will not be rendered and thus not take up vertical space, no?

src/registry/children/ChildrenTable.tsx
```diff
@@ -0,0 +1,193 @@
...
+                  aria-label={intl.formatMessage(
+                    {
+                      description: "Children component: child 'selected' aria label",
+                      defaultMessage: 'Include {firstname}?',
+                    },
+                    {
+                      firstname: child.firstNames,
+                    }
+                  )}
+                  onChange={onSelectionChanged}
```
sergei-maertens: shouldn't we use `onBlur` instead of `onChange` for consistency? Every other field is marked as touched when the user moves focus away, not when the value changed.

The idea is: the user has interacted with the field, so it is now touched, so now we can display any validation errors that may exist instead of overwhelming them with validation errors before they've even touched the field.

src/components/forms/Checkbox/Checkbox.tsx
```diff
@@ -56,6 +56,10 @@ export interface CheckboxProps {
    * assist users in filling out the field correctly.
    */
   tooltip?: React.ReactNode;
+  /**
+   * Callback when the checkbox value changes.
+   */
+  onChange?: (e: React.ChangeEvent<HTMLInputElement>) => void;
```
sergei-maertens: this is unused now, right? please remove it

src/registry/children/BareCheckbox.tsx
robinmolen: For regular checkboxes we set the default value to `false` in the getInitialValues (or whatever the component configuration decides is the default value). The server-fetched children data that we receive only contain personal information (bsn, firstnames, date of birth), so `selected` is undefined.

Formik uses regular html input behaviour for the checkboxes. When a input (including checkboxes) has no value, the value `"on"` is apparently used (https://html.spec.whatwg.org/multipage/input.html#checkbox-state-(type=checkbox), https://html.spec.whatwg.org/multipage/input.html#dom-input-value-default-on)

So setting the default value to `False` ensures that the input knows which values the checkbox should have (`true` or `false`).

And for some reason, Formik treats this checkbox as a checkboxes group (which weirdly enough doesn't happen with the default value set?). So when checking the checkbox, for server-fetched children, the value is set to `["on"]`.

robinmolen: I'll expend the comment to convey this clearer

sergei-maertens: isn't the right fix to ensure `selected` is set to `false` or `undefined` in `getInitialValues` then? And/or a `useEffect` hook at the component level to set the value if it should be set (because of visible/hidden toggles)

src/registry/children/Children.scss
sergei-maertens: and why can't we apply a grid/flexbox row gap for the children component? :thinking:

PR #208, opened 2025-11-13
[#77] Support displaying custom validation error messages

### LINKED ISSUE #77: Support displaying custom validation error messages
For a number of validation error types (they all have a key in Formio like `required`, `minDate` etc.), the form designer can supply there own error message to be displayed. This feature is currently not supported.

**Tasks**

- [x] For each component type, ensure that we have a type-safe way of accessing the custom error message if provided
- [ ] The error messages support some interpolation - figure out how to integrate with react-intl, as i18next has a different interpolation format.
- [x] Ensure that the default validation errors match what is currently defined in Formio/SDK/our backend (see `formio/{en,nl}.json` in the backend)

vaszig:
Closes #77 partly

- This PR handles only the first subtask of the ticket (making sure that custom error messages are accessed by the components that support this functionality)

### DIFF
--- a/src/registry/date/validationSchema.ts
+++ b/src/registry/date/validationSchema.ts
@@ -24,31 +24,26 @@ const getValidationSchema: GetValidationSchema<DateComponentSchema> = (
   const maxDate = datePicker?.maxDate;
 
   let dateSchema = z.coerce.date();
-
   if (minDate) {
     dateSchema = dateSchema.min(parseISO(minDate), {message: errors?.minDate});
   }
-
   if (maxDate) {
     dateSchema = dateSchema.max(parseISO(maxDate), {message: errors?.maxDate});
   }
 
-  let stringSchema: z.ZodFirstPartySchemaTypes = z.string({required_error: errors?.required});
-
-  stringSchema = stringSchema.refine(
-    value => {
-      if (!value && !required) return true;
-
-      const parsed = parseISO(value);
-      return isValid(parsed);
-    },
-    {message: errors?.invalid_date || intl.formatMessage(INVALID_INPUT_MESSAGE)}
-  );
-
-  let schema: z.ZodFirstPartySchemaTypes = stringSchema.pipe(dateSchema);
+  let schema: z.ZodFirstPartySchemaTypes = z
+    .string({required_error: errors?.required})
+    .refine(
+      value => {
+        const parsed = parseISO(value);
+        return isValid(parsed);
+      },
+      {message: errors?.invalid_date || intl.formatMessage(INVALID_INPUT_MESSAGE)}
+    )
+    .pipe(dateSchema);
 
   if (!required) {
-    schema = z.union([schema, z.literal(''), z.undefined()]);
+    schema = schema.optional().or(z.literal(''));
   }
 
   if (plugins.length) {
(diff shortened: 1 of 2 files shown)


src/registry/textfield/validationSchema.ts
```diff
@@ -7,18 +7,27 @@ const getValidationSchema: GetValidationSchema<TextFieldComponentSchema> = (
...
 ) => {
-  const {key, validate = {}, multiple} = componentDefinition;
+  const {key, validate = {}, multiple, errors} = componentDefinition;
   const {required, maxLength, pattern, plugins = []} = validate;
 
-  let schema: z.ZodFirstPartySchemaTypes = z.string();
-  if (maxLength !== undefined) schema = schema.max(maxLength);
+  let schema: z.ZodFirstPartySchemaTypes = z.string(
+    errors?.required ? {required_error: errors.required} : undefined
+  );
```
sergei-maertens:
```suggestion
  let schema: z.ZodFirstPartySchemaTypes = z.string({required_error: errors?.required});
```
tests still pass if you do it this way and apply this pattern everywhere - the types allow for `required_error` / `message` itself to be undefined, so we can use that to short-circuit and avoid duplication.

src/registry/textfield/validationSchema.ts
```diff
@@ -7,18 +7,27 @@ const getValidationSchema: GetValidationSchema<TextFieldComponentSchema> = (
...
+
+  if (maxLength !== undefined) {
+    schema = errors?.maxLength
+      ? schema.max(maxLength, {message: errors.maxLength})
+      : schema.max(maxLength);
+  }
```
sergei-maertens:
```suggestion
    schema = schema.max(maxLength, {message: errors?.maxLength});
  }
```
the most important part is `schema.max(...)` here, as it encodes the actual validation rule. having to duplicate that is error-prone, because an implementation change or refactor requires you to take into account all occurrences - that's bound to result in bugs.

Instead, the types allow for `{message: undefined}`, so we can use that :-) 

I see one possible problem where the key is present in errors and has an empty string value, but I'd classify that as a bug in the backend rather than the frontend code.

src/registry/iban/validationSchema.ts
```diff
@@ -46,7 +46,7 @@ const getValidationSchema: GetValidationSchema<IbanComponentSchema> = (
   if (multiple) {
     schema = z.array(schema);
     if (required) {
-      schema = schema.min(1);
+      schema = schema.min(1, errors?.required ? {message: errors.required} : undefined);
```
sergei-maertens:
```suggestion
      schema = schema.min(1, {message: errors?.required});
```

src/registry/phoneNumber/validationSchema.ts
```diff
@@ -50,7 +55,7 @@ const getValidationSchema: GetValidationSchema<PhoneNumberComponentSchema> = (
   if (multiple) {
     schema = z.array(schema);
     if (required) {
-      schema = schema.min(1);
+      schema = schema.min(1, errors?.required ? {message: errors.required} : undefined);
```
sergei-maertens:
```suggestion
      schema = schema.min(1, {message: errors?.required});
```

sergei-maertens (changes requested):
Note - I haven't tested any of these suggestions so you may be doing these deliberately because of type checker things, please confirm if that's the case :)

PR #213, opened 2025-11-24
Implement customer profile component

### LINKED ISSUE #178: Implement component type: profile
Implement the new component type `profile` (internal technical name will be `customerProfile`), this should be part of the "special" category. The component is in active development, so the specifications might change.

The component contains 1 piece of configuration that has effect on the frontend: Digital address channels (naming might change). "Digital address channels" is a radio button configuration field, which dictates which digital address types are shown in the component.

#### Digital address channels options
There should always be at least one digital address type.
- Email address only: Only the email address field is shown.
- Phone number only: Only the phone number field is shown.
- Email address and phone number: Show inputs for email address and phone number.

#### Field inputs
The email and phoneNumber field inputs change based on prefilled data:
- If there is prefilled digital addresses, the fields should use select inputs with the prefilled digital addresses as options and the preferred digital address as default value. The select inputs should allow the defining of new digital addresses.
- If there are no prefilled digital addresses, then text inputs should be used.

#### To implement
- [x] Digital addresses fetching mechanism
- [x] Add new digital address, when there are already digital addresses known (do we use a creatable select input?)
- [x] Validation
- [x] Digital address decision flow: use digital address for this case only, or as new preferred digital address
- [ ] When adding a new digital address, you should not be able to enter an already known digital address

#### Questions
- If there are prefilled digital addresses, and the UI input is a select field, how does the user add a new digital address?
  Do we use some sort of creatable select input? Or do we add a button, which chances the select field to a text field?
  - If we go the creatable select input route, we cannot disable the select input when there is only one prefilled digital address. (Just something to keep in mind)
- If we have prefilled digital addresses, but none of the addresses are marked as preferred, what is then the default value of the select field? Just the first digital address? Or is there then no default value?

It was decided to postpone the digital address verify. This has impact on the "how do you want to use the digital address" choices:
- Should we still use a non-closable modal, to force the user to choose? (And allow easy implementation of verify functionality)
- How do we trigger the showing of the choice fields, when using a text field?
  Do we add a button to trigger the modal, or show the choice fields below the digital address input when the input value changes?
- How do we trigger the showing of the choice fields, when using a select field?
  Do we add a button to trigger the modal, or is the modal opened when a different digital address is selected? Or, if we no-longer go the modal route, do we show the choice fields below the digital address input when a different digital address is selected?

See design: https://www.figma.com/design/wgX6HfrX93VI2ScQnVzxbK/OF---Open-Klant?node-id=0-1&p=f&t=I3ebtA6PxlVzfK7r-0

robinmolen:
Closes #178

Implementing the customer profile component (publicly this should be known as the "Profile" component).

The customer profile component can fetch digital addresses for the communication preferences prefill plugin. This will result in a list of addresses and preferences, grouped by digital address type. This PR contains a basic setup for this behaviour, which will be expanded/fixed after the new communication preferences API is designed and implemented: https://github.com/open-formulieren/open-forms/issues/5772.

The profile component currently supports digital addresses of the types: 'email' and 'phoneNumber'. If other digital address types should be supported, the types library and this renderer implementation will have to be updated.

TODO:
~- [ ] Ensure that a new digital address isn't already known~ This will be moved to the backend, and will be treated as a valid situation (i.e. won't raise validation errors, and will server the user as best as possible)

## Profile component states
The component can start in two states:
1. Some digital addresses are known of the user
2. No digital addresses are known of the user

### Some digital addresses are known of the user
When there are known digital addresses, we show a select input. By default, we select the preferred digital address, and otherwise the first digital address we retrieved. Below the select input, we show a "add new digital address" button. When this button is click, the select input will be replaced with a textfield.

The "creating a new digital address" follows the "No digital addresses are known of the user" UI and logic, with one additional rule: you cannot enter an already known digital address (as we don't want to save the same digital addresses twice for the same user).

### No digital addresses are known of the user
It could be no digital addresses are known, or could be fetched, for the user. This could be because the user isn't logged in, isn't known in the customer interactions service, or doesn't have any known digital addresses of the expected type in the customer interactions service.

In this scenario, we show a textfield for the digital address type. When a valid digital address is entered, a "update preferences" button will appear. The "update preferences" button will open a modal, in which the user can decide how this digital address should be used. By default the digital address will only be used for the current form they are filling in. This modal will, in the future, be used to perform the digital address verification.

### DIFF
--- a/src/registry/customerProfile/initialValues.ts
+++ b/src/registry/customerProfile/initialValues.ts
@@ -1,19 +1,43 @@
-import type {CustomerProfileComponentSchema, CustomerProfileData} from '@open-formulieren/types';
+import type {
+  CustomerProfileComponentSchema,
+  CustomerProfileData,
+  DigitalAddress,
+} from '@open-formulieren/types';
 
 import type {GetInitialValues} from '@/registry/types';
+
+function assertNotArrayOfArray(
+  value: CustomerProfileData | CustomerProfileData[]
+): asserts value is CustomerProfileData {
+  // test if any item inside the array is an Array, indicating nested arrays
+  if (value.some(child => Array.isArray(child))) {
+    throw new TypeError('An array of arrays is not supported and should not happen.');
+  }
+}
 
 const getInitialValues: GetInitialValues<
   CustomerProfileComponentSchema,
   CustomerProfileData | null
-> = ({key, defaultValue, digitalAddressTypes}: CustomerProfileComponentSchema) => {
+> = ({
+  key,
+  defaultValue = [] satisfies CustomerProfileData,
+  digitalAddressTypes,
+}: CustomerProfileComponentSchema) => {
+  // side-effect of the generic types in formio components, but realistically we don't
+  // expect any defaultValue to ever be set.
+  assertNotArrayOfArray(defaultValue);
+
   if (defaultValue === undefined || !defaultValue?.length) {
-    defaultValue = digitalAddressTypes.map(type => ({
-      type,
-      address: '',
-      preferenceUpdate: 'useOnlyOnce',
-    }));
+    defaultValue = digitalAddressTypes.map(
+      type =>
+        ({
+          type,
+          address: '',
+          preferenceUpdate: 'useOnlyOnce',
+        }) satisfies DigitalAddress
+    );
   }
-  return {[key]: defaultValue as unknown as CustomerProfileData};
+  return {[key]: defaultValue};
 };
 
 export default getInitialValues;


src/registry/customerProfile/digitalAddresses.ts
```diff
@@ -0,0 +1,69 @@
...
+    if (!result) {
+      // Initial the form with empty values if fetch failed.
+      digitalAddressTypes.forEach((type, index) => {
+        setFieldValue(`${profileComponentKey}.${index}`, {
+          address: '',
+          type: type,
+          useOnlyOnce: true,
+        } satisfies DigitalAddress);
+      });
+      setDigitalAddresses([]);
```
sergei-maertens: you can also just make this the return state of the `useAsync` hook, it already uses `useState` under the hood so you're basically duplicating the implementation and using only half of the hook :)

```ts
const {loading, value, error} = useAsync(...);
// value contains the return value from the wrapped callback
```

src/registry/customerProfile/digitalAddresses.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    const addressesResultMap: Partial<Record<DigitalAddressType, DigitalAddressGroup>> =
+      Object.fromEntries(result.map(addressGroup => [addressGroup.type, addressGroup]));
+
+    digitalAddressTypes.forEach((type, index) => {
+      const addressData: DigitalAddressGroup | undefined = addressesResultMap[type];
+      // The default value is the preferred address or the first address in the list.
+      const defaultAddress = !addressData
+        ? ''
+        : (addressData.preferred ?? addressData.addresses[0] ?? '');
```
sergei-maertens:
```suggestion
      const defaultAddress = addressData?.preferred ?? addressData?.addresses?.[0] ?? '';
```
achieves the same thing, I think?

src/registry/customerProfile/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+  componentDefinition: CustomerProfileComponentSchema;
+}
+
+export const FormioCustomerProfile: React.FC<FormioCustomerProfileProps> = ({
+  componentDefinition: {key: name, label, tooltip, description, validate, digitalAddressTypes},
+}) => {
+  const {getFieldMeta} = useFormikContext();
+  const id = useId();
+  name = useFieldConfig(name);
+  const {touched, error: formikError} = getFieldMeta(name);
```
sergei-maertens: I'm not sure about using this `touched` state, because the email/phone number fields inside (indexed access) will also have an individual touched state.

So a touched with `{profile: true}` would be overwritten with `{touched: [true, false]}` by virtue of index access, I think.

Can we infer the field-as-a-whole touched state from the individual items touched state? And clicking the button to update preferences would then mark that item as touched too - I think that covers all possible situations.

src/registry/customerProfile/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+          {label}
+          {tooltip && <Tooltip>{tooltip}</Tooltip>}
+        </>
+      }
+      isInvalid={invalid}
+      hasTooltip={!!tooltip}
+      aria-describedby={[descriptionId, errorMessageId].filter(Boolean).join(' ')}
+      className="openforms-customer-profile"
+    >
+      <div className="openforms-customer-profile__form-field-container">
```
sergei-maertens: why not use the `FormFieldContainer` component as wrapper?


```suggestion
      <FormFieldContainer>
```

sergei-maertens: I'm making a note that this is wrong in the addressNL component too 🤔 will follow up on that

src/registry/customerProfile/prefill.stories.ts
```diff
@@ -0,0 +1,253 @@
...
+    // Wait for the prefill to finish
+    await waitFor(() => {
+      // Both email and phone number fields are displayed as dropdowns
+      expect(emailField).toHaveRole('combobox');
+      expect(phoneNumberField).toHaveRole('combobox');
+    });
+
+    // The fetched addresses are displayed in the dropdowns
+    expect(rsValue(emailField)).toBe('foo@test.com');
+    expect(rsValue(phoneNumberField)).toBe('0612345678');
```
sergei-maertens: iirc, react-select does have proper label + hidden input linking so you should just be able to do `expect(canvas.getByLabelText(...)).toHaveValue(...)` - is there a reason this doesn't work here?

And in another PR I added an option to render a submit button and you should be able to pass an `onSubmit` parameter, so you I'd much rather prefer we test what the submission data looks like instead of DOM inspection:

```ts
expect(context.parameters.formik.onSubmit).toHaveBeenCalledWith({customerProfile: [...]})
```

src/registry/customerProfile/subFields.tsx
```diff
@@ -0,0 +1,332 @@
+import type {DigitalAddress, DigitalAddressType} from '@open-formulieren/types';
+import {ButtonGroup} from '@utrecht/button-group-react';
+import {Alert} from '@utrecht/component-library-react';
```
sergei-maertens: are the necessary styles loaded for this component?

src/registry/customerProfile/subFields.tsx
```diff
@@ -0,0 +1,332 @@
...
+}
+
+const DigitalAddressTextfield: React.FC<DigitalAddressTextfieldProps> = ({
+  type,
+  namePrefix,
+  fieldName,
+  isRequired,
+  textfieldProps,
+}) => {
+  const {setFieldValue, getFieldMeta} = useFormikContext();
```
sergei-maertens: here too you can probably use the generic type trick for better type checking support

src/registry/customerProfile/subFields.tsx
```diff
@@ -0,0 +1,332 @@
...
+  const [isModalOpen, setIsModalOpen] = useState(false);
+
+  const {value, touched, error} = getFieldMeta(fieldName);
+
+  // Only show the preference button if the field is touched, has a value and has no error
+  const showPreferencesButton = touched && !error && value !== '';
```
sergei-maertens: this negates part of my earlier remark about setting touched state when this button is clicked - I think this behaviour makes most sense

src/registry/customerProfile/subFields.tsx
```diff
@@ -0,0 +1,332 @@
...
+              description="Profile digital address new address: default preference text"
+              defaultMessage={`The {digitalAddressType, select,
+                email {email address}
+                phoneNumber {phone number}
+                other {{digitalAddressType}}
+              } you provide will automatically only be used for this form.
+              You can update your preferences using the button below.`}
+              values={{digitalAddressType: type}}
+            />
+          </Alert>
```
sergei-maertens: I am a bit confused if this should be fed as `description` to the text field rather than using an info alert. I'm inclined to go with the description.

src/registry/customerProfile/subFields.tsx
```diff
@@ -0,0 +1,332 @@
...
+ * Only for new digital addresses can the user update their preferences. Updating the
+ * preference can be done in a modal, once the digital address textfield contains a valid
+ * digital address. In other situations, the user will have to use the online portal
+ * (outside of Open Forms) to update their preferences.
+ *
+ * @param namePrefix The name prefix for the fields.
+ * @param type The type of digital address field.
+ * @param isRequired Whether the fields are required.
+ * @param digitalAddressGroup The digital address group to display.
+ * @param textfieldProps Additional props to pass to the text input.
```
sergei-maertens: I notice we're ending up with two documentation styles - the majority uses the docblock style in the interfaces/types:

```ts
interface Foo {
  /**
   * A bar, not the kind where you can order drinks.
   */
  bar: string;
}
```

I definitely prefer this form because the documentation is super close to the actual code/type it documents, whereas the JSDoc style has greater risk of being outdated and too detached.

Both get picked up in MDX, I believe.

src/registry/customerProfile/subFields.tsx
```diff
@@ -0,0 +1,332 @@
...
+  useEffect(() => {
+    setUseTextInput(!hasAddresses);
+  }, [hasAddresses]);
+
+  return (
+    <Fieldset
+      isInvalid={invalid}
+      aria-describedby={errorMessageId}
+      className="openforms-customer-profile__digital-address-fieldset"
+    >
```
sergei-maertens: This is a fieldset for only a single input field, no? Does this need to be a fieldset? It has no legend, no aria-describedby that's meaningful (except for the error message).

src/registry/customerProfile/digitalAddressPreferencesModal.tsx
```diff
@@ -0,0 +1,113 @@
...
+        }}
+      >
+        <Form>
+          <FormFieldContainer>
+            <BareRadioField
+              name="preference"
+              options={[
+                {
+                  value: 'isNewPreferred',
+                  label: (
```
sergei-maertens: I see some more type-safety potential here once you define the `FormValues` 😄

src/registry/customerProfile/validationSchema.ts
```diff
@@ -0,0 +1,172 @@
...
+ * The generic rules for digital addresses are defined in this builder. These include:
+ * - The `type` field must be one of the allowed types for the digital address.
+ * - The `address` is optional or required based on the component `required` parameter.
+ *
+ * @param required - `required` parameter from the parent component.
+ * @param type - For which type of digital address the schema is meant.
+ * @param addressSchema - The validation schema for the address field.
+ */
+const buildDigitalAddressSchema = (
+  required: boolean,
```
sergei-maertens: we tend to prefix booleans with `is` or `has`


```suggestion
  isRequired: boolean,
```

it reads very natural in JSX when used as bare prop: `<Component isRequired />`, but in general it's a good naming pattern.

src/registry/customerProfile/validationSchema.ts
```diff
@@ -0,0 +1,172 @@
...
+  description:
+    'Validation error for customerProfile with multiple digital addresses for the same type',
+  defaultMessage: 'You cannot submit multiple digital addresses for the type {digitalAddressType}.',
+});
+
+const EMAIL_INVALID_MESSAGE = defineMessage({
+  description:
+    'Validation error for customerProfile digital address that does not match the email format',
+  defaultMessage: 'Invalid email address.',
+});
```
sergei-maertens: So I'd expect the message for an invalid email to be identical if it's for the `email` or `customerProfile` component, which probably motivates defining this message string centrally and re-use it in both places, and give it a description that is decoupled from either of those components.

Otherwise in the translation catalog we need to ensure the same translations are used, and that's going to fail at some point, guaranteed.

`src/i18n/messages.ts` looks like a good candidate, where you use `defineMessage(s)` for the shared translations.

src/registry/customerProfile/validationSchema.ts
```diff
@@ -0,0 +1,172 @@
...
+    schema = schema.length(digitalAddressTypes.length);
+  } else {
+    schema = schema.optional();
+  }
+
+  // We have to do the superRefine on the top-level instead of on the individual schemas.
+  // The discriminated union schema does not allow ZodEffects, only ZodObjects, so the
+  // individual digital address schemas must be plain ZodObjects.
+  schema = schema.superRefine((items: DigitalAddress[] | null, ctx) => {
+    if (items == null) return;
```
sergei-maertens: I don't see a `z.null()` anywhere - how can this be `null`?

src/registry/customerProfile/BareRadioField.tsx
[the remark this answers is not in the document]
robinmolen: I added this as a separate component because of the situation we had with the checkboxes for the children component. Making the label on the radio button optional would cause the same issue the checkboxes had: you remove the explicit functionality, and potentially lose functionality with the tooltip (where do we place the tooltip, if there is no visible label?)

I think it would be best to add an `aria-description`, or something similar to the `BareRadioField`.

src/registry/customerProfile/digitalAddresses.ts
```diff
@@ -0,0 +1,69 @@
...
+import {useFormSettings} from '@/hooks';
+
+import type {DigitalAddressGroup, DigitalAddressesResponseBody} from './types';
+
+interface UseDigitalAddresses {
+  digitalAddresses: DigitalAddressesResponseBody;
+}
+
+export const useDigitalAddresses = (
+  profileComponentKey: string,
```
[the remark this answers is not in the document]
robinmolen: I thought `profileComponentKey` would make it more explicit that we expect the `key` (or `name`) of the profile component. Maybe something like `profileComponentName` then?

src/registry/customerProfile/digitalAddresses.ts
```diff
@@ -0,0 +1,69 @@
...
+        } satisfies DigitalAddress);
+      });
+      setDigitalAddresses([]);
+      return;
+    }
+
+    const addressesResultMap: Partial<Record<DigitalAddressType, DigitalAddressGroup>> =
+      Object.fromEntries(result.map(addressGroup => [addressGroup.type, addressGroup]));
+
+    digitalAddressTypes.forEach((type, index) => {
```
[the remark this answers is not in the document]
robinmolen: Creating a map outside the `digitalAddressTypes.forEach` means we only loop twice. Once on the `result` and once on the `digitalAddressTypes`.

I don't think we should trust the `result` 100% with regards of which digital addresses we should support (things can break). So thats why im looping over the `digitalAddressTypes`.

I can implement it with a `result.find`, thats fine. The increase in processing cost is minimal. I just thought, "if we can make it cheaper with a simple map, then why not do it"?

src/registry/customerProfile/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const {getFieldMeta} = useFormikContext();
+  const id = useId();
+  name = useFieldConfig(name);
+  const {touched, error: formikError} = getFieldMeta(name);
+  const {digitalAddresses} = useDigitalAddresses(name, digitalAddressTypes);
+
+  const error = formikError as unknown as
+    | undefined
+    | string
+    | (string | FormikErrors<DigitalAddress>)[];
```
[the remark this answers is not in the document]
robinmolen: Mhh, i'll take a look if I can find a solution for all of these cases.

But maybe this is better done outside this PR? To prevent blocking

src/registry/customerProfile/subFields.tsx
```diff
@@ -0,0 +1,332 @@
...
+ */
+const getDigitalAddressOptions = (
+  intl: IntlShape,
+  digitalAddressGroup: DigitalAddressGroup
+): Option[] =>
+  digitalAddressGroup.addresses.map(address => ({
+    label: address,
+    value: address,
+    description:
+      !!digitalAddressGroup.preferred && address === digitalAddressGroup.preferred
```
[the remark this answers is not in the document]
robinmolen: I added that because `preferred` could be undefined. But `address` cannot (or should not) be undefined, so `address === digitalAddressGroup.preferred` is indeed enough

src/registry/customerProfile/subFields.tsx
robinmolen: Will move/update it

src/registry/customerProfile/validationSchema.spec.ts
```diff
@@ -0,0 +1,320 @@
...
+  shouldUpdateCustomerData: false,
+  digitalAddressTypes: ['email', 'phoneNumber'],
+};
+
+const buildValidationSchema = (component: CustomerProfileComponentSchema) => {
+  const schemas = getValidationSchema(component, {
+    intl,
+    getRegistryEntry,
+    validatePlugins: async (plugins: string[]) =>
+      plugins.includes('fail') ? 'not valid' : undefined,
```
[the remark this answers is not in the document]
robinmolen: Ah no its not. Updating it

src/registry/customerProfile/digitalAddressPreferencesModal.tsx
```diff
@@ -0,0 +1,113 @@
...
+                    <FormattedMessage
+                      description="Profile digital address preferences modal 'isNewPreferred' option label"
+                      defaultMessage={`Save my data for the future forms.
+                        You can edit your preferences in the online <a>portal</a>.`}
+                      values={{
+                        a: (...chunks) => (
+                          <a href={portalUrl} target="_blank" rel="noopener noreferrer">
+                            {chunks}
+                          </a>
```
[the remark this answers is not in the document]
robinmolen: Ahh yeah i like that idea.. The `useCustomerProfileComponentParameters` you earlier suggested might be it bit of overkill then. I'll take a look if that is still needed/nice

src/registry/customerProfile/validationSchema.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  let schema: z.ZodFirstPartySchemaTypes = z.array(
+    z.discriminatedUnion('type', digitalAddressSchemas)
+  );
+
+  if (required) {
+    // When required, a value for each digital address type must be provided
+    schema = schema.length(digitalAddressTypes.length);
+  } else {
+    schema = schema.optional();
```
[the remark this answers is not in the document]
robinmolen: Ah good point. Yeah and you cannot actually remove a digital address item (decrease the array size). You can, at most, assign an empty address value.

Removing the `.optional()` bit

src/registry/customerProfile/index.tsx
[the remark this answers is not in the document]
robinmolen: Adding cleanup ticket. Leaving this for now

src/registry/customerProfile/subFields.tsx
robinmolen: Moved it to description 👍

src/registry/customerProfile/validationSchema.ts
```diff
@@ -0,0 +1,172 @@
...
+import type {IntlShape} from 'react-intl';
+import {defineMessage} from 'react-intl';
+import {z} from 'zod';
+
+import type {GetValidationSchema} from '@/registry/types';
+
+const USE_ONLY_ONCE_AND_AS_PREFERRED_CONSTRAINT_MESSAGE = defineMessage({
+  description:
+    'Validation error for customerProfile digital address useOnlyOnce and isNewPreferred that are both true',
+  defaultMessage: "The digital address cannot be marked as both 'one-time use' and 'preferred'.",
```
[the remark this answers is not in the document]
robinmolen: Like discussed in office: New types version, new profile preferences structure 🥳

sergei-maertens (approved):

PR #214, opened 2025-11-24
[#102] Formio map component

### LINKED ISSUE #102: Implement component type: map

SonnyBA:
Closes #102 

Adds the FormIO side of the Map component, see https://github.com/open-formulieren/formio-renderer/pull/184 for the component that is used in this PR.

### DIFF
--- a/src/reference/clearOnHide.stories.ts
+++ b/src/reference/clearOnHide.stories.ts
@@ -474,7 +474,7 @@ const {custom: DependentFields, reference: DependentFieldsReference} = storyFact
 
     await userEvent.type(field1, 'hidden', {delay: 200});
 
-    const field5 = await canvas.findByLabelText('Field 5');
+    const field5 = await canvas.findByLabelText('Field 5', undefined, {timeout: 500});
     expect(field5).toBeVisible();
 
     await userEvent.click(canvas.getByRole('button', {name: 'Submit'}));


src/registry/map/validationSchema.ts
```diff
@@ -0,0 +1,64 @@
...
+    z.object({
+      type: z.literal('LineString'),
+      coordinates: lineValue,
+    }),
+    z.object({
+      type: z.literal('Polygon'),
+      coordinates: polygonValue,
+    }),
+    z.null(),
+  ]);
```
sergei-maertens: this can be made more strict, and then should also make the value parsing error handling obsolete in https://github.com/open-formulieren/formio-renderer/pull/214/files#diff-a8399940fa4c9f7a4d69aaf85be8dce8c4455d5ddc686e024603223e64493a3dR81 - which I suspect was only added to be able to introduce an invalid state with non-sense data.

You can inspect `componentDefinition.interactions` to check which shapes are allowed and compose the schema that way. Also, looking at `validate.required` determines whether null is allowed or not.

I'm adding a commit to implement gradual schema building, you can then set up a story that dissallows a particular type but provide valid geojson to trigger the error :)

PR #216, opened 2025-11-25
Editgrid: item level validation and fix cancel button

### LINKED ISSUE #36: Implement repeating group (editgrid)

sergei-maertens:
Closes #36
Closes open-formulieren/open-forms#3661

The formio.js variant blocks form (step) submission if there are editgrids with 'draft' items. An item is in 'draft' state when it's expaneded after clicking the 'edit button' or when it's freshly added. To take it out of draft state, the "save" or "cancel" button need to be clicked.

Additionally, in draft/edit mode, the delete button is a mistake, instead it should be a cancel button that restores the original state
of the item.

Note that this pollutes the item data, but it's the only way to get this marker in there. On submit of the draft, the marker is removed, so the server should never see this. The first attempt to use a `Symbol` for this failed because there's no way to make this play nice with Zod, see https://github.com/colinhacks/zod/issues/2734

### DIFF
--- a/src/components/forms/EditGrid/EditGrid.tsx
+++ b/src/components/forms/EditGrid/EditGrid.tsx
@@ -3,6 +3,7 @@ import {FormField} from '@utrecht/component-library-react';
 import {FieldArray, getIn, setIn, useFormikContext} from 'formik';
 import type {FormikErrors} from 'formik';
 import {useId, useMemo} from 'react';
+import {flushSync} from 'react-dom';
 import {FormattedMessage} from 'react-intl';
 import type {ValidationError} from 'zod-formik-adapter';
 
@@ -130,7 +131,7 @@ function EditGrid<T extends {[K in keyof T]: JSONValue} = JSONObject>({
   ...props
 }: EditGridProps<T>) {
   name = useFieldConfig(name);
-  const {getFieldProps, errors, getFieldHelpers} = useFormikContext();
+  const {getFieldProps, errors, getFieldHelpers, validateField} = useFormikContext();
   const {value: formikItems} = getFieldProps<T[] | undefined>(name);
   const id = useId();
   const labelId = label ? `${id}-label` : undefined;
@@ -195,20 +196,32 @@ function EditGrid<T extends {[K in keyof T]: JSONValue} = JSONObject>({
                     getItemBody={props.getItemBody}
                     canEditItem={props.canEditItem}
                     saveItemLabel={props.saveItemLabel}
-                    onChange={(newValue: T) => {
-                      arrayHelpers.replace(index, newValue);
-                      // clear any (initial) errors for this item - since this callback
-                      // is invoked, it implies that the schema validation passed. Any
-                      // external (backend) errors will require re-validation of the whole
-                      // form by the backend, so we can safely clear those backend errors.
-                      if (!hasFieldLevelError) {
-                        const {setError} = getFieldHelpers(`${name}.${index}`);
-                        setError(undefined);
-                      }
+                    onChange={async (newValue: T) => {
+                      // force the Formik state update so that the validate call sees
+                      // the up-to-date state for a new validation run
+                      flushSync(() => {
+                        arrayHelpers.replace(index, newValue);
+                        // clear any (initial) errors for this item - since this callback
+                        // is invoked, it implies that the schema validation passed. Any
+                        // external (backend) errors will require re-validation of the whole
+                        // form by the backend, so we can safely clear those backend errors.
+                        if (!hasFieldLevelError) {
+                          const {setError} = getFieldHelpers(`${name}.${index}`);
+                          setError(undefined);
+                        }
+                      });
+                      await validateField(name);
                     }}
                     canRemoveItem={canRemoveItem}
                     removeItemLabel={removeItemLabel}
-                    onRemove={() => arrayHelpers.remove(index)}
+                    onRemove={async () => {
+                      // force the Formik state update so that the validate call sees
+                      // the up-to-date state for a new validation run
+                      flushSync(() => {
+                        arrayHelpers.remove(index);
+                      });
+                      await validateField(name);
+                    }}
                     validate={props.validate}
                   />
                 );
--- a/src/registry/editgrid/index.stories.ts
+++ b/src/registry/editgrid/index.stories.ts
@@ -6,7 +6,7 @@ import type {
   FieldsetComponentSchema,
 } from '@open-formulieren/types';
 import type {Meta, StoryObj} from '@storybook/react-vite';
-import {expect, fn, userEvent, within} from 'storybook/test';
+import {expect, fn, userEvent, waitFor, within} from 'storybook/test';
 
 import FormioComponent from '@/components/FormioComponent';
 import type {FormioFormProps} from '@/components/FormioForm';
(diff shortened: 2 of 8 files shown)


src/components/forms/EditGrid/EditGrid.tsx
```diff
@@ -125,19 +132,37 @@ function EditGrid<T extends {[K in keyof T]: JSONValue} = JSONObject>({
...
       )}
       <FieldArray name={name} validateOnChange={false}>
         {arrayHelpers => (
           <div className="openforms-editgrid">
             {/* Render each item wrapped in an EditGridItem */}
-            <ol className="openforms-editgrid__container">
+            <ol
+              className="openforms-editgrid__container"
+              aria-labelledby={labelId}
+              aria-describedby={errorMessageId}
```
robinmolen: Lets also add an id for the description field, and include that in the `aria-describedby`

src/components/forms/EditGrid/EditGridItemButtons.tsx
```diff
@@ -89,9 +87,18 @@ const IsolationModeButtons: React.FC<IsolationModeButtonsProps> = ({
...
-      )}
+
+      <PrimaryActionButton hint="danger" onClick={onCancel} aria-describedby={ariaDescribedBy}>
+        {cancelLabel || (
+          <FormattedMessage
+            // yes, this description looks off, but that's actually the formio.js meaning
+            // even though the actual behaviour is 'cancel' and not remove (unless it's
+            // a freshly added item)
+            description="Edit grid item default remove button label"
+            defaultMessage="Cancel"
```
robinmolen: I don't get this comment. The description is used to identify translatable texts, right? So we know when and where its used, so we can define/provide correct translations (and if we ever give users the ability to provide their own translation files, this will cause confusion. Because: is this the "cancel" remove button, or is this the "remove" remove button?).

So its better to accurately describe this text, then to listen to the meaning of formio.js, right? Because, this has no impact on formio.js?

Also, won't this clash with the exact same description used for the default remove button label? Does the i18n code understand that these are two different texts?

src/registry/editgrid/index.stories.ts
```diff
@@ -629,3 +633,101 @@ export const WithNestedRadio: Story = {
...
+  play: async ({canvasElement, context, step}) => {
+    const canvas = within(canvasElement);
+
+    await userEvent.click(canvas.getByRole('button', {name: 'Submit'}));
+    expect(await canvas.findByText('Array must contain at least 1 element(s)')).toBeVisible();
+    expect(context.args.onSubmit).not.toHaveBeenCalled();
+
+    await step('Verify item error display', async () => {
+      await userEvent.click(canvas.getByRole('button', {name: 'Add another'}));
+    });
```
robinmolen: Is there something missing here? Should something happen after the button is clicked?

Looking at the storybook test; I guess the "Array must contain at least 1 element(s)" error should be removed, right?

src/registry/editgrid/validationSchema.ts
```diff
@@ -0,0 +1,52 @@
...
+  const {required} = validate;
+
+  let schema: z.ZodFirstPartySchemaTypes = z.array(
+    // editgrid is always an array of objects. The object themselves are validated
+    // when the item is saved (it's a nested Formik state with its own validation
+    // schema).
+    // TODO if we allow inline editing -> the full schema needs to be provided, and you
+    // need to take into account the field visibility based on the data & clearOnHide
+    // behaviour for each individual item, so that'll probably need to be done in a
+    // `superRefine` call because not every item has the same schema!
```
robinmolen: Won't a discriminatedUnion work for the different component schemas? Like a more complex setup of the profile component validation?

Either way, sounds like a hoot 😃

src/components/forms/EditGrid/EditGrid.tsx
```diff
@@ -191,10 +217,13 @@ function EditGrid<T extends {[K in keyof T]: JSONValue} = JSONObject>({
...
-                      setIndexToAutoExpand(formikItems.length);
-                    }
-                    arrayHelpers.push(emptyItem);
+                    let newItem: MarkedEditGridItem<T> = setIn(
+                      emptyItem,
+                      ITEM_EXPANDED_MARKER,
+                      true
+                    );
+                    newItem = setIn(newItem, ITEM_ADDED_MARKER, true);
+                    arrayHelpers.push(newItem);
```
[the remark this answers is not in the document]
sergei-maertens: Yes, because `setIn` keeps the object identity whereas the object spreading approach creates new references of objects through the shallow copy.

Our editgrid handling/hooks operate on changes in formik data to calculate the new validation schemas, with a lot of `useMemo` applied to prevent infinite re-renders. Not using `setIn` will lead to such bugs because they break memoization.

src/components/forms/EditGrid/EditGridItemButtons.tsx
```diff
@@ -89,9 +87,18 @@ const IsolationModeButtons: React.FC<IsolationModeButtonsProps> = ({
...
+          />
+        )}
+      </PrimaryActionButton>
```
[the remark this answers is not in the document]
sergei-maertens: Will check if this can be simplified, there was some conflicting usage of `RemoveButton` iirc but maybe I didn't pay enough attention to the specifics.

src/components/forms/EditGrid/EditGridItemButtons.tsx
sergei-maertens: > Because: is this the "cancel" remove button, or is this the "remove" remove button?).

That's why the comment is added explaining why it looks off: they're the same in Formio. Both the cancel and remove button have the "cancel" label, but the behaviour is different depending on the item state.

> Also, won't this clash with the exact same description used for the default remove button label? Does the i18n code understand that these are two different texts?

Yesn't - the combination of `defaultMessage` and `description` is the input for the message ID.

I'll take a look if we can correct this or not.

src/registry/editgrid/index.tsx
```diff
@@ -218,7 +219,7 @@ export const FormioEditGrid: React.FC<EditGridProps> = ({
   // ensure we keep setting a deeper scope when dealing with nesting
   const parentScope = !isRoot ? setIn(grandParentValues, keyPrefix, parentValues) : parentValues;
 
-  const initialValues = extractInitialValues(components, getRegistryEntry);
+  const initialValues = deepMergeValues(extractInitialValues(components, getRegistryEntry), {});
```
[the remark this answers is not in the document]
sergei-maertens: `foo.bar` didn't result in `{foo: {bar: ...}}`, but in `{'foo.bar': ...}` instead.

Tests failed because of the broken behaviour, so it shouldn't be accidentally revert, but I'll add a comment.

src/components/forms/EditGrid/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const ITEM_EXPANDED_MARKER = '_OF_INTERNAL_isItemExpanded';
+export const ITEM_ADDED_MARKER = '_OF_INTERNAL_isNewItem';
```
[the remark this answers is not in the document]
sergei-maertens: Created #224

PR #220, opened 2025-11-26
Fix error validation that's too eager (radio, date, datetime with pickers)

### LINKED ISSUE #218: The 'touched' + validation error state is triggered too early in certain circumstances
> validation is triggered very quickly (especially noticeable with keyboard navigation on radio where it even flickers when switching between values, also makes users "misclick" the value they intend to choose. Also interacting with datepicker) 

similar case:

> Keyboard navigation in radio to select value incorrectly blurs/validates too soon and doesn't clear the error when a value is selected (tab into empty radio, no value selected, arrow-down to select second option -> validation error displays)

- [x] Fix focus/touched state on radio keyboard navigation (for required radio)
- [x] Prevent date/datetime validation errors from being displayed while the calendar picker modal is open

sergei-maertens:
Closes #218

I'm noticing some more less-than-ideal situations with the datepicker and datetime picker w/r to focus management, but don't want to blow up the scope and want to see if they come up during testing :-)

### DIFF
--- a/src/components/forms/DateTimeField/DateTimeField.tsx
+++ b/src/components/forms/DateTimeField/DateTimeField.tsx
@@ -291,6 +291,13 @@ const DateTimeField: React.FC<DateTimeFieldProps> = ({
           onChange={onPartChange}
           value={time}
           className="utrecht-textbox--openforms"
+          onKeyDown={async (event: React.KeyboardEvent) => {
+            // treat enter key as a submit
+            if (event.key === 'Enter') {
+              setIsOpen(false);
+              await validateField(name);
+            }
+          }}
         />
       </FloatingWidget>
       <HelpText>{description}</HelpText>


src/components/forms/DateField/DatePicker/DateField-DatePicker.stories.tsx
```diff
@@ -266,3 +267,41 @@ export const NoAsterisksForRequired = {
...
   },
 };
+
+// don't display validation errors if there's still focus on a radio input
+export const NoErrorWhileFocus: Story = {
+  decorators: [withMockDate],
+  args: {
+    widget: 'datePicker',
+    name: 'date',
+    label: 'No error displayed while keyboard navigation happens',
```
viktorvanwijk: I guess it's not specifically about keyboard navigation here, but more about the picker still having focus, right?

sergei-maertens:
> > I'm noticing some more less-than-ideal situations with the datepicker and datetime picker w/r to focus management, but don't want to blow up the scope and want to see if they come up during testing :-)
> 
> Perhaps you discovered this as well, but I had already noticed before that if you type a date(time) in the textbox, and immediately click on it to open the date picker, the value is not updated yet. Only after the field loses focus, the picker is updated.

ah that's a new one, I'll update the parent issue!

PR #223, opened 2025-11-27
[#221] Disable checkboxes when max allowed is reached

### LINKED ISSUE #221: Checkboxes with max only show validation error, does not disable checkboxes when max is reached

vaszig:
Closes #221

Sometimes it's needed to disable a checkbox for some reason. For example the when a max limit of allowed checked checkboxes is reached we disable the rest of them. The custom error message for the maxSelectedCount has not been removed because we may have an edge case were it will be shown and useful.

### DIFF
--- a/src/registry/selectboxes/index.stories.tsx
+++ b/src/registry/selectboxes/index.stories.tsx
@@ -208,6 +208,40 @@ export const ValidateRequired: ValidationStory = {
     ).toBeVisible();
   },
 };
+
+export const MaxAllowedReached: ValidationStory = {
+  ...BaseValidationStory,
+  args: {
+    onSubmit: fn(),
+    componentDefinition: {
+      id: 'component1',
+      type: 'selectboxes',
+      key: 'my.selectboxes',
+      label: 'A selectboxes field',
+      defaultValue: {option1: false, option2: false, option3: false},
+      values: [
+        {value: 'option1', label: 'Option 1'},
+        {value: 'option2', label: 'Option 2'},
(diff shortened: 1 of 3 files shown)


src/registry/selectboxes/index.stories.tsx
```diff
@@ -208,6 +208,40 @@ export const ValidateRequired: ValidationStory = {
...
+export const MaxAllowedReached: ValidationStory = {
+  ...BaseValidationStory,
+  args: {
+    onSubmit: fn(),
+    componentDefinition: {
+      id: 'component1',
+      type: 'selectboxes',
+      key: 'my.selectboxes',
+      label: 'A selectboxes field',
+      defaultValue: {option1: false, option2: false},
```
sergei-maertens:
```suggestion
      defaultValue: {option1: false, option2: false, option3: false},
```

src/registry/selectboxes/index.tsx
```diff
@@ -50,6 +50,20 @@ export const FormioSelectboxes: React.FC<FormioSelectboxesProps> = ({componentDe
 
     const {value} = getFieldProps(key);
 
+    // Disable rest of the checkboxes if maxSelectedCount has been reached
+    if (maxSelectedCount) {
```
sergei-maertens: this is a good example of where synchronizing state with an effect is an anti pattern - we already have existing state/information to figure out if/what needs to be disabled, so the entire additional `useState` is not necessary.

I've refactored this, take a look so you see what I mean :-)

PR #233, opened 2025-12-02
[#225] Fix value not updated in date and datetime (date picker)

### LINKED ISSUE #225: Fix value not updated when typing in the Date picker textbox
If you type a date(time) in the textbox, and immediately click on it to open the date picker, the value is not updated yet. Only after the field loses focus, the picker is updated.

vaszig:
Closes #225

### DIFF
--- a/src/components/forms/DateField/DatePicker/DateField-DatePicker.stories.tsx
+++ b/src/components/forms/DateField/DatePicker/DateField-DatePicker.stories.tsx
@@ -306,3 +306,37 @@ export const NoErrorWhileFocus: Story = {
     expect(input).not.toHaveFocus();
   },
 };
+
+export const TypedDateAndDatePickerUpdate: Story = {
+  args: {
+    widget: 'datePicker',
+    name: 'date',
+    label: 'Date',
+    isDisabled: false,
+    isRequired: false,
+  },
+  parameters: {
+    formik: {
+      onSubmit: fn(),
+    },
+  },
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+    const widget = canvas.queryByRole('dialog');
+
+    expect(widget).toBeNull();
+    const date = canvas.getByLabelText('Date');
+    await userEvent.type(date, '29-08-2025');
+    expect(date).toHaveDisplayValue('29-08-2025');
+
+    // Ensure that the date is properly highlighted/shown in the calendar without
+    // loosing focus
+    await userEvent.click(date);
+    expect(await canvas.findByRole('dialog')).toBeVisible();
+    const selectedEventButton = await canvas.findByRole('button', {
+      name: 'vrijdag 29 augustus 2025',
+    });
+    expect(selectedEventButton).toBeVisible();
+    expect(selectedEventButton).toHaveClass('utrecht-calendar__table-days-item-day--selected');
+  },
+};
--- a/src/components/forms/DateField/DatePicker/DatePicker.tsx
+++ b/src/components/forms/DateField/DatePicker/DatePicker.tsx
@@ -1,7 +1,7 @@
 import {Paragraph, Textbox} from '@utrecht/component-library-react';
 import {formatISO} from 'date-fns';
 import {useField, useFormikContext} from 'formik';
-import {useId} from 'react';
+import {useId, useState} from 'react';
 import {flushSync} from 'react-dom';
 import {useIntl} from 'react-intl';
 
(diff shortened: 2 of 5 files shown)


src/components/forms/DateField/DatePicker/DatePicker.tsx
```diff
@@ -139,17 +129,39 @@ const DatePicker: React.FC<DatePickerProps> = ({
...
+              : rawValue
+          }
+          onChange={event => {
+            // Set the raw value in the textbox and update the formik value depending on
+            // if we have a valid date or not
+            const newRaw = event.target.value;
+            setRawValue(newRaw);
+            const parsed = parseDate(newRaw, dateLocaleMeta);
+            setValue(parsed ? formatISO(parsed, {representation: 'date'}) : rawValue);
+          }}
```
viktorvanwijk: I guess one option that we didn't try last week is to move the (original) value parsing from the `onBlur` handler to the `onChange` handler. That way the parsed value is set to the field immediately, and the picker is updated as well.

This does mean that the value will be formatted directly as you're typing, which might not be ideal? :thinking:

src/registry/date/index.stories.tsx
```diff
@@ -105,6 +105,45 @@ export const WithTooltipDatePicker: Story = {
   },
 };
 
+export const TypedDateAndDatePickerUpdate: Story = {
```
viktorvanwijk: Any specific reason why you added the test here, and not in `DateField-DatePicker.stories.tsx`?

viktorvanwijk (commented):
Some thoughts :)

There currently is a bug for both components when you enter an invalid date(time), blur the field to initiate validation, then clear the entire field and blur again. The validation error is still visible even though the field is cleared. Seems like there might be a mismatch between the textbox value and the field value?

src/components/forms/DateField/DatePicker/DatePicker.tsx
viktorvanwijk: > by adding an extra zero at the end and not being able to edit it after the onBlur event

Seems like similar issues to what we saw last week then :(

src/reference/clearOnHide.stories.ts
```diff
@@ -474,7 +474,7 @@ const {custom: DependentFields, reference: DependentFieldsReference} = storyFact
 
     await userEvent.type(field1, 'hidden', {delay: 200});
 
-    const field5 = await canvas.findByLabelText('Field 5', undefined, {timeout: 500});
+    const field5 = await canvas.findByLabelText('Field 5', undefined, {timeout: 1000});
```
vaszig: This was changed to `1000` in order to make CI happy.

src/components/forms/DateField/DatePicker/DateField-DatePicker.stories.tsx
```diff
@@ -306,3 +306,38 @@ export const NoErrorWhileFocus: Story = {
...
+    name: 'date',
+    label: 'Date',
+    isDisabled: false,
+    isRequired: false,
+  },
+  parameters: {
+    formik: {
+      onSubmit: fn(),
+    },
+    chromatic: {disableSnapshot: true}, // don't create snapshots because we can't set the timezone for chromatic
```
viktorvanwijk: This should not be relevant for the date field

src/components/forms/DateField/DatePicker/DatePicker.tsx
```diff
@@ -139,17 +138,25 @@ const DatePicker: React.FC<DatePickerProps> = ({
...
-            const date = parseDate(value, dateLocaleMeta);
-            // If we were able to create a date object, format it to an ISO-8601 string and set it
-            // as the field value. Otherwise, just set the entered value to the field directly.
-            // It's up to the validation libraries to check it.
-            const newValue = date ? formatISO(date, {representation: 'date'}) : value;
-            await setValue(newValue);
+            const parsed = parseDate(rawValue, dateLocaleMeta);
+            if (parsed) {
+              // We have a date object, format it according to the locale (and remove
+              // comma) and use as the textbox value.
```
viktorvanwijk:
```suggestion
              // We have a date object, format it according to the locale and use as the textbox value.
```
Probably copied over from the datetime component?

viktorvanwijk (commented):
Just some minor things :)

sergei-maertens:
Closing this - see the issue [comment](https://github.com/open-formulieren/formio-renderer/issues/225#issuecomment-3617053423) for details.

PR #248, opened 2025-12-09
[#186] Make Empty state validation consistent for all the components

### LINKED ISSUE #186: Consistent empty state validation
This ticket is derived from a PR discussion: https://github.com/open-formulieren/formio-renderer/pull/152#discussion_r2464824595.

We should follow a consistent ruling for marking components as empty or non-empty. This ruling should be: a component is empty, as long as its field, or fields, are empty. 

## Array fields
The empty state for array fields should be as followed:
| Component value | Empty state |
|-----------------|-------------|
| `undefined`     | Empty       |
| `null`          | Empty       |
| `[]`            | Empty       |
| `['']`           | Empty   |
| `['bar']`           | Non-empty   |

## Object fields
The empty state for object fields should be as followed:
| Component value | Empty state |
|-----------------|-------------|
| `undefined`     | Empty       |
| `null`          | Empty       |
| `{}`            | Empty       |
| `{'foo': ''}`           | Empty   |
| `{'foo': 'bar'}`           | Non-empty   |

## Complex fields / Editgrid component
The empty state for more complex fields (mixing array and object values), like the editgrid component, should be as followed:
| Component value | Empty state |
|-----------------|-------------|
| `undefined`     | Empty       |
| `null`          | Empty       |
| `[]`            | Empty       |
| `[{}]`            | Empty       |
| `[{'foo': ''}]`           | Empty   |
| `[{'foo': 'bar'}]`           | Non-empty   |

Each component should be checked if this logic is correctly applied, and should be update if they don't. For the editgrid component it's known that it doesn't follow these rules correctly, and should be updated.

To facilitate the this ruling for editgrid components, the soft required component would potentially need to validate nested values, which it currently (probably) doesn't. This will also need to updated/implemented.

vaszig:
Closes #186

### DIFF
--- a/src/registry/editgrid/empty.spec.ts
+++ b/src/registry/editgrid/empty.spec.ts
@@ -1,6 +1,11 @@
-import type {EditGridComponentSchema, TextFieldComponentSchema} from '@open-formulieren/types';
+import type {
+  EditGridComponentSchema,
+  SelectboxesComponentSchema,
+  TextFieldComponentSchema,
+} from '@open-formulieren/types';
 import {expect, test} from 'vitest';
 
+import {getRegistryEntry} from '@/registry';
 import type {JSONObject} from '@/types';
 
 import isEmpty from './empty';
@@ -8,10 +13,31 @@ import isEmpty from './empty';
 test.each([
   // Empty states
   [undefined, true],
+  [null, true],
   [[], true],
+  [[{}], true],
+  [[{}, {}], true],
+  [[{textfield: ''}], true],
+  [[{textfield: ''}, {}], true],
+  [[{textfield: ''}, {textfield1: ['']}], true],
+  [[{textfield1: []}], true],
+  [[{selectboxes: {}}], true],
+  [[{selectboxes: {'123': false}}], true],
+  [[{textfield: '', textfield1: [], selectboxes: {}, editgrid1: [{textfield2: ''}]}], true],
   // Non-empty states
-  [[{}], false],
-  [[{textfield: ''}], false],
+  [[{textfield: 'test'}], false],
+  [[{editgrid1: [{textfield2: 'foo'}]}], false],
+  [
+    [
+      {
+        textfield: '',
+        textfield1: [],
+        selectboxes: {'123': false},
+        editgrid1: [{textfield2: 'foo'}],
+      },
+    ],
+    false,
+  ],
 ])(
   'Editgrid isEmpty compares against defined, non-empty array state of value',
   (valueToTest: JSONObject[] | undefined, expected: boolean) => {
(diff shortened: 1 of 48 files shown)


src/registry/editgrid/empty.spec.ts
```diff
@@ -29,10 +37,32 @@ test.each([
...
+          disableAddingRemovingRows: false,
+          components: [
+            {
+              type: 'textfield',
+              key: 'textfield2',
+              id: 'textfield2',
+              label: 'Textfield2',
+            } satisfies TextFieldComponentSchema,
+          ],
+        } satisfies EditGridComponentSchema,
```
robinmolen: The components and testcases don't fully overlap. The `selectboxes` is missing, and `textfield` is used as multiple while not defined as such.

Could you add `selectboxes` to the components? And mark `textfield1` as `multiple=true`, and adjust the test data?

src/registry/editgrid/empty.spec.ts
```diff
@@ -1,17 +1,25 @@
...
   [undefined, true],
+  [null, true],
   [[], true],
+  [[{}], true],
+  [[{}, {}], true],
+  [[{textfield: ''}], true],
+  [[{textfield: ''}, {textfield1: ''}], true],
+  [[{textfield: []}], true],
+  [[{selectboxes: {}}], true],
+  [[{textfield: '', textfield1: '', selectboxes: [], editgrid1: [{textfield2: ''}]}], true],
```
robinmolen:
```suggestion
  [[{textfield: '', textfield1: '', selectboxes: {}, editgrid1: [{textfield2: ''}]}], true],
```

src/registry/editgrid/empty.ts
```diff
@@ -1,13 +1,53 @@
...
+ * are considered as empty or not.
+ *
+ * Each (child) component is validated according to its `isEmpty` implemented function.
+ *
+ * @param parentComponentDefinition The parent (editgrid) component definition.
+ * @param componentValue The value of the editgrid component.
+ *
+ * @return The function returns whether the (child) component is empty (true) or not (false).
+ */
+export function isEmptyChildComponent(
```
robinmolen:
```suggestion
export function isEmptyChild(
```
As the function validates the editgrid child as a whole, not an individual editgrid child component.

A name like `hasEmptyChildComponent` could also work

src/registry/editgrid/empty.ts
```diff
@@ -1,13 +1,53 @@
...
+  for (const childrenKey of childrenKeys) {
+    const value = componentValue[childrenKey];
+    const childDefinition = parentComponentDefinition.components.find(c => c.key === childrenKey);
+
+    if (!childDefinition) continue;
+
+    const registry = getRegistryEntry(childDefinition);
+    if (!registry?.isEmpty) continue;
+
+    const result = registry.isEmpty(childDefinition, value, getRegistryEntry);
```
robinmolen:
```suggestion
    const isChildEmpty = registry.isEmpty(childDefinition, value, getRegistryEntry);
```
Its a small detail, but I think its best to keep this function as clear as possible. (Just looking at myself, keeping track of the double negatives is already difficult enough)

src/registry/editgrid/empty.ts
```diff
@@ -1,13 +1,53 @@
...
+
+const isEmpty: IsEmpty<EditGridComponentSchema, JSONObject[]> = (
+  _componentDefinition,
+  value,
+  getRegistryEntry
+) => {
+  if (value == null || value.length === 0 || value.every(obj => Object.keys(obj).length === 0))
+    return true;
+
+  return value.every(val => isEmptyChildComponent(_componentDefinition, val, getRegistryEntry));
```
robinmolen:
```suggestion
  if (value == null || value.length === 0) return true;

  return value.every(val => isEmptyChildComponent(_componentDefinition, val, getRegistryEntry));
```
I think it would be best to validate all child data using the `isEmptyChildComponent`. Otherwise we create inconsistencies in the empty validation, which will result in inconsistent results.

(I noticed this while trying to figure why the `editgrid/empty.spec.ts` `[[{}, {}], true]` succeeds, but `[[{textfield: ''}, {}], true]` would fail)

src/registry/editgrid/empty.spec.ts
vaszig: I left selectboxes out on purpose (component definition not found). But if you think it should be there I can add it, of course. And for the textfield you are right, I missed that.

robinmolen: Im not sure I understand your meaning of "component definition not found". Was that an error thrown by the renderer? Because `selectboxes` should work..?

PR #249, opened 2025-12-09
Profile component portal url

robinmolen:
Partly closes https://github.com/open-formulieren/open-forms/issues/5795

In the profile component we show multiple messages stating that users can update their communication preferences using an online portal. The url of this portal can be configured in the OF backend. Because this url could be empty, we need to make sure we only mention the portal when there is an actual portal to mention. Otherwise, we show dead links and cause confusion for users.

### DIFF
--- a/src/registry/customerProfile/index.stories.tsx
+++ b/src/registry/customerProfile/index.stories.tsx
@@ -87,7 +87,7 @@ export const OpenPreferencesModal: Story = {
 
       // Enter an email address
       await userEvent.type(emailField, 'test@mail.com');
-      await emailField.blur();
+      emailField.blur();
 
       // The "update preferences" button should now be visible
       await userEvent.click(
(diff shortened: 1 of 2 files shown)


src/components/forms/Select/Select.tsx
```diff
@@ -171,10 +172,10 @@ const Select: React.FC<SelectProps> = ({
           await setTouched(true);
           await validateField(name);
         }}
-        aria-describedby={errorMessageId}
+        aria-describedby={[descriptionId, errorMessageId].filter(Boolean).join(' ')}
```
sergei-maertens: this makes the description linking to the input more inconsistent, since anything that's `TextField` based does *not* have this behaviour.

Please revert this and keep it out of scope for this ticket, and solve it for *all* the components in one go in a separate ticket.

I do understand the intent, but I feel that inconsistency is worse than not having this at this point (the legacy renderer also does not have this and it has never come up in a11y audits)

sergei-maertens (changes requested):

PR #254, opened 2025-12-15
Fix dependencies in package.json to prevent node_modules in dist/

sergei-maertens:
This is probably messing up the dependency resolution, breaking the SDK build.


sergei-maertens:
@SonnyBA this seems to resolve the issues when using the components in the SDK, tested with `npm pack` and a local install + storybook. I could reproduce the various issues and resolve them.

PR #258, opened 2025-12-17
Profile component moving between form-steps resets user input

robinmolen:
Closes https://github.com/open-formulieren/open-forms/issues/5829

Moving between form-steps would cause the digital addresses to be fetched again, which would reset the user input. In addition, the subfield usage logic didn't take into account a user entering a new digital address and moving between form steps.

Both bugs are solved by these commits.

**Changes:**
1. The default address is only set if there isn't already an address value
2. Which subfield is used is now based on two things:
    - Are there server-fetched addresses?
    - Is the current address part of the server-fetched addresses list?
3. The "which subfield is used" logic is swapped around and renamed. Going from double negative logic, to a simpler positive logic.

### DIFF
--- a/src/registry/customerProfile/index.stories.tsx
+++ b/src/registry/customerProfile/index.stories.tsx
@@ -238,6 +238,9 @@ export const WithDigitalAddressValidationError: Story = {
     expect(
       await canvas.findByText('Generic error message from digital address validation.')
     ).toBeVisible();
+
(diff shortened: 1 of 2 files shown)


sergei-maertens:
Can you rebase on main? I merged the other PR.

sergei-maertens (approved):
Code looks good :)

PR #260, opened 2025-12-18
Introducing profile component parameter `updatePreferencesModalEnabled`

robinmolen:
Closes https://github.com/open-formulieren/open-forms/issues/5824

Using the `updatePreferencesModalEnabled` we can control whether the "update preferences" button and modal will be shown along the profile component textfield subfield.

In addition, the logic behind when the "update preferences" button appears is also updated. The `touched` state caused issues in OF when moving between form-steps, where the "update preferences" button would not be shown because the field wasn't `touched`. To make the "update preferences" button appear, `updatePreferencesModalEnabled` needs to be set to `true`, and the textfield needs an valid digital address.

### DIFF
--- a/src/registry/customerProfile/index.stories.tsx
+++ b/src/registry/customerProfile/index.stories.tsx
@@ -194,6 +194,40 @@ export const OpenPreferencesModalWithoutPortalUrl: Story = {
   },
 };
 
+export const WithUpdatePreferencesModalEnabledSetToFalse: Story = {
+  args: {
+    componentDefinition: {
+      id: 'customerProfile',
+      type: 'customerProfile',
+      key: 'customerProfile',
+      label: 'Profile',
+      description:
+        'The "update preferences" button should never be visible because `updatePreferencesModalEnabled` is set to `false`.',
+      digitalAddressTypes: ['email'],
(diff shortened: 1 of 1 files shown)


src/registry/customerProfile/index.stories.tsx
```diff
@@ -158,6 +160,45 @@ export const OpenPreferencesModalWithoutPortalUrl: Story = {
...
+    const emailField = await canvas.findByLabelText('Email');
+
+    // Enter an email address
+    await userEvent.type(emailField, 'test@mail.com');
+    emailField.blur();
+
+    // The "update preferences" button should not be visible
+    await waitFor(() => {
+      expect(canvas.queryByRole('button', {name: 'Update preferences'})).not.toBeInTheDocument();
+    });
```
sergei-maertens: These assertions are hard to get right - the assert can pass for many different (wrong) reasons:

* the button is there, but with a different label
* the component is async and momentarily doesn't show it, but does show it after 500ms for whatever reason

So I don't trust these kind of assertions and tend to leave them out myself, instead relying on a good story name, comment in the test and visual check in Chromatic that no unexpected elements are there.

Long story to suggest removing this assertion because IMO it's too fragile.

PR #269, opened 2025-12-19
Fix date and datetime fields marked as readonly

### LINKED ISSUE #265: Check how the date/datetime fields behaviour when they're marked as readonly
* Textinput itself should be disabled/readonly
* The calendar/datepicker should not be possible to open, I suspect this not to be the case now and would allow setting values which is not intended.

Verify & fix if indeed broken.

viktorvanwijk:
Closes #265

Changes:
* Ensured the datepicker trigger icon is disabled when the field is disabled


vaszig (commented):
Tested locally and now works as expected for both components. I have a question because I am not so familiar with the frontend.

PR #277, opened 2025-12-23
Validate datepicker when focusout

### LINKED ISSUE #274: Date/datetime fields trigger (required) validation too soon
When the fields are marked as required and you <kbd>Tab</kbd> navigate to the field and then the date picker trigger, the field validation kicks on because of the `TextField` blur and results in the "Field is required" error being displayed while focus is still "within" the field.

Probably we need an approach here like the radio/selectboxes to track focus and only validate when focus is left completely.

**Components to fix**:
- [x] Date
- [x] Datetime

robinmolen:
Closes #274

The date-picker validation should only trigger when the focus is outside the component. Using a similar setup as the radio field, by adding a container element, and click and focus listeners, we can validate when the date-picker has focus or not.

### DIFF
--- a/src/components/forms/DateField/DatePicker/DatePicker.tsx
+++ b/src/components/forms/DateField/DatePicker/DatePicker.tsx
@@ -125,7 +125,7 @@ const DatePickerField: React.FC<DatePickerFieldProps> = ({
         {label}
       </Label>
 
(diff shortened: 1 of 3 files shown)


sergei-maertens (approved):
I expected more pain, nice!

PR #307, opened 2026-04-07
Replace testing infrastructure

### LINKED ISSUE #306: Replace storybook test runner & testing-library with vitest browser mode equivalents
@storybook/test-runner is deprecated in favour of tapping into vitest browser mode, while testing-library is also being phased out in favour of built-in equivalents

sergei-maertens:
Closes #306

* Upgrade to vitest 4
* Switched unit tests to vitest browser mode
* Upgrade to Storybook 10
* Added & configured `@storybook/vitest-addon`
* Removed `@storybook/test-runner`
* Replaced testing-library usage in unittest with vitest-browser/vitest-browser-react
* Removed testing-library dependencies


src/registry/file/File.spec.tsx
```diff
@@ -298,43 +270,39 @@ test('process file rejection eagerly', async () => {
...
-  ).toBeVisible();
-  expect(screen.getByText(/The file must be smaller than 1 MB\./)).toBeVisible();
-  expect(screen.getByText(/The file is too small\./)).toBeVisible();
-  expect(screen.getByText(/Not ignored\./)).toBeVisible();
+  await expect
+    .element(screen.getByText(/The uploaded file type is not an allowed type. It must be \.pdf\./))
+    .toBeVisible();
+  await expect.element(screen.getByText(/The file must be smaller than 1 MB\./)).toBeVisible();
+  await expect.element(screen.getByText(/The file is too small\./)).toBeVisible();
+  await expect.element(screen.getByText(/Not ignored\./)).toBeVisible();
```
annashamray, line 294: Could you please explain why you added `await` here? I thought that `getByText` is synchronous

src/registry/file/File.spec.tsx
```diff
@@ -151,140 +150,113 @@ const TestComponent: React.FC<TestComponentProps> = ({
...
   // initially, no uploads are displayed
-  const uploadsList = screen.queryAllByRole('listitem');
+  const uploadsList = screen.getByRole('listitem').all();
   expect(uploadsList).toHaveLength(0);
 
-  await user.click(screen.getByRole('button', {name: 'Simulate upload'}));
+  await screen.getByRole('button', {name: 'Simulate upload'}).click();
 
-  const pendingUploadsList = screen.queryAllByRole('listitem');
+  const pendingUploadsList = screen.getByRole('listitem').all();
```
annashamray, line 218: Why do you replace `queryAllByRole` here? Is it not recommended to use anymore?

src/components/FormioForm.spec.tsx
```diff
@@ -428,13 +424,13 @@ describe('Updating form errors', () => {
       },
     });
 
-    expect(await screen.findByText('Visible foo error')).toBeVisible();
-    expect(await screen.findByText('Visible bar.baz error')).toBeVisible();
+    await expect.element(screen.getByText('Visible foo error')).toBeVisible();
+    await expect.element(screen.getByText('Visible bar.baz error')).toBeVisible();
```
annashamray, line 428: Could you also explain please why you replaced `findByText` with `await + getByText` here?

annashamray (approved):
Looks good! 
I just have a couple of questions too better understand testing best practices

src/registry/file/File.spec.tsx
sergei-maertens: will be clarified in the lighting talk, but with vitest-browser, the `expect.element(...)` is async because it uses locators rather than elements!

src/registry/file/File.spec.tsx
sergei-maertens: Indeed, it's a difference between testing-library & vitest-browser mode. Locators vs. elements :-)

src/components/FormioForm.spec.tsx
sergei-maertens: Same reason as above!

src/registry/file/File.spec.tsx
annashamray: The LT timing is perfect then! :)

PR #310, opened 2026-04-10
Switch to individual NL DS packages

### LINKED ISSUE #301: Upgrade to latest versions of Utrecht React components
We're quite behind on packages used.

**Tasks**

- [x] Upgrade to latest @utrecht/component-library-react
- [x] Swap out @utrecht/component-library-react with the individual packages
- [x] Remove `--utrecht-button-icon-gap` from design tokens package for our theme

**Downstream tasks**

- [x] Update packages + CSS used in backend public stylesheet
- [x] Updated packages + CSS used in SDK repo

Everything needs to be aligned to avoid conflicting/duplicated CSS.

sergei-maertens:
Closes #301

Install and use the individual React component packages instead of using the mono/aggregation package.


eslint.config.mjs
```diff
@@ -38,6 +38,31 @@ const config = defineConfig([
...
+                'PrimaryActionButton',
+                'Button',
+                'SecondaryActionButton',
+                'SubtleButton',
+              ],
+              message: 'Use Button (variants) from `@/components/Button` instead.',
+            },
+            {
+              name: '@utrecht/component-library-react',
+              importNames: [
```
annashamray, line 53: Won't you have import error here, because this package is removed from dependencies?

sergei-maertens: it's not yet removed because not all components are available yet as standalone package!

annashamray (approved):

PR #312, opened 2026-04-16
Fix crash with soft required file in editgrid made visible with logic

### LINKED ISSUE #6185: Soft-required file component inside hidden editgrid causes form crash (new renderer)
More specifically when the editgrid is made visible by either conditional or backend logic.

To reproduce:
1. Create form with checkbox, soft-required errors, and editgrid (hidden by default) with soft-required file component inside
2. Add logic rule that shows the editgrid when the checkbox is checked
3. Fill in the form and check the checkbox
4. Observe crash :boom:

viktorvanwijk:
Closes open-formulieren/open-forms#6185

### DIFF
--- a/src/components/FormioForm.stories.tsx
+++ b/src/components/FormioForm.stories.tsx
@@ -739,3 +739,61 @@ export const WithMultipleAndDefaultValue: Story = {
     });
   },
 };
+
+export const SoftRequiredFileComponentInsideEditgridWithLogic: Story = {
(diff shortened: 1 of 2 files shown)


robinmolen (approved):
Looks great!

viktorvanwijk:
I think this needs backporting to 3.4.x and 3.5.x, but forgot to add the label to the issue. Effectively meaning releasing a new patch version and updating the stable branches in the SDK

PR #314, opened 2026-04-17
Update package-lock.json

viktorvanwijk:
This is the result of running `npm i` locally. When I tried this yesterday for the release, CI was complaining about `fsevents` not being installed, see [action](https://github.com/open-formulieren/formio-renderer/actions/runs/24512615185/job/71803803451). So I manually updated just the version numbers to 1.5.2 in `package-lock.json` and released that.

But when trying to update the renderer in the SDK, I'm getting dependency conflicts, and I'm not entirely sure yet if they are caused by this `package-lock.json` file not being updated properly. EDIT: doesn't seem to be the case, there is a conflict with `vega` being on version 6.2.0, while vega is defined as "^5.25.0" in @utrecht/component-library-react@13.1.0

CI seems to accept it now, though :thinking:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -201,7 +201,6 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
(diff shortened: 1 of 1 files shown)


viktorvanwijk:
Closing this because it won't be relevant after releasing 1.5.3 (#316)

sergei-maertens:
Upstream utrecht is also working on extracting the vega-related component/package into a standalone package so hopefully this goes through in the next few days and cleans up this mess

PR #315, opened 2026-04-20
Fix poor UX with date(time) field validation

### LINKED ISSUE #6163: new renderer: date validation not triggering on blur/change

3.5.0

### Customer reference

_No response_

The date fields do validate on change/blur. instead the trigger to validate seems to be different for different types of validation 

Not required, not (date) validation: Triggers on blur. 
Image

Required, no (date) validation: does not trigger on blur, but does trigger when interacting with another component in the step. 
Image

Required, with extra date validation (in future/past): Does not trigger on blur, Does not validate when interacting with other components. But they do trigger when interacting with a datepicker (any!) in the same step.

Image
Image

One more scenario:
when date field is validated against a different datefield (in future) and picking a date that is valid, the error will not go away (unless interacting with the other date field):

Image
Image

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

viktorvanwijk:
Closes open-formulieren/open-forms#6163

Note that the ticket only describes the date field, but the datetime field has the same problem.

### DIFF
--- a/src/components/forms/DateField/DatePicker/DateField-DatePicker.stories.tsx
+++ b/src/components/forms/DateField/DatePicker/DateField-DatePicker.stories.tsx
@@ -418,3 +418,34 @@ export const ReadOnlyDatePicker: Story = {
     expect(trigger).toHaveAttribute('tabIndex', '-1');
   },
 };
+
+export const ValidateOnBlur: Story = {
+  args: {
+    name: 'validateOnBlur',
+    label: 'Validate on blur',
+    isRequired: false,
+  },
+  parameters: {
+    formik: {
+      initialValues: {
+        validateOnBlur: '',
+      },
+      zodSchema: z.object({
+        validateOnBlur: z.any().refine(() => false, {message: 'Always invalid'}),
+      }),
+    },
+  },
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+
+    const date = canvas.getByLabelText('Validate on blur');
+    expect(date).not.toHaveAttribute('aria-invalid');
+
+    await userEvent.type(date, '20-04-2026');
+    expect(date).toHaveFocus();
+
+    date.blur();
+    expect(await canvas.findByText('Always invalid')).toBeVisible();
+    expect(date).toHaveAttribute('aria-invalid', 'true');
+  },
+};
--- a/src/components/forms/DatePicker.tsx
+++ b/src/components/forms/DatePicker.tsx
@@ -201,46 +201,26 @@ export const DatePickerRoot: React.FC<DatePickerRootProps> = ({
     const container = containerRef.current;
     if (!container || !onFocusOut) return;
 
-    const handleClick = (event: MouseEvent) => {
-      const clickedElement = event.target;
-      // ignore clicking inputs or labels, as they result in focus events which we
-      // already listen to.
-      // Clicking within the container (like clicking the datepicker calendar trigger)
-      // should also not trigger validation.
-      if (
-        clickedElement instanceof HTMLLabelElement ||
-        clickedElement instanceof HTMLInputElement ||
-        container.contains(clickedElement as Node)
-      )
-        return;
-
-      // any other element (anywhere) being clicked triggers validation, as focus loss
-      // is implied
-      onFocusOut();
-    };
-
     /**
-     * Detect focus events on anything other than elements inside the div container.
+     * Handle focus out events.
      *
-     * Receiving this event means that some node got focus - if it's outside of our
-     * container we can trigger validation because we're 100% certain that the container
-     * does not have focus.
+     * Receiving this event means that some node lost focus. We don't have to trigger validation
+     * if the element is not part of the container or if the new target node is still within
+     * the container.
      */
-    const handleFocusIn = (event: FocusEvent) => {
-      const focusedElement = event.target;
-      if (container.contains(focusedElement as Node)) return;
+    const handleFocusOut = (event: FocusEvent) => {
+      const currentTarget = event.target;
+      const newTarget = event.relatedTarget;
+      if (!container.contains(currentTarget as Node) || container.contains(newTarget as Node))
+        return;
       onFocusOut();
     };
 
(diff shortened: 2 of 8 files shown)


src/components/forms/DatePicker.tsx
```diff
@@ -201,46 +201,26 @@ export const DatePickerRoot: React.FC<DatePickerRootProps> = ({
...
      */
-    const handleFocusIn = (event: FocusEvent) => {
-      const focusedElement = event.target;
-      if (container.contains(focusedElement as Node)) return;
+    const handleOnBlur = (event: FocusEvent) => {
+      const blurredElement = event.target;
+      const newTarget = event.relatedTarget;
+      if (!container.contains(blurredElement as Node) || container.contains(newTarget as Node))
+        return;
       onFocusOut();
```
robinmolen, line 216: I would rename the `DatePickerRoot` function `onFocusOut` to `onBlur`, to more accurately describe when its called.

src/components/forms/DatePicker.tsx
```diff
@@ -201,46 +201,26 @@ export const DatePickerRoot: React.FC<DatePickerRootProps> = ({
...
+     * the container.
      */
-    const handleFocusIn = (event: FocusEvent) => {
-      const focusedElement = event.target;
-      if (container.contains(focusedElement as Node)) return;
+    const handleOnBlur = (event: FocusEvent) => {
+      const blurredElement = event.target;
+      const newTarget = event.relatedTarget;
+      if (!container.contains(blurredElement as Node) || container.contains(newTarget as Node))
+        return;
```
robinmolen: The blur validation trigger seems to be fired prematurely when using the datepicker menu.

When you open the datepicker menu using the trigger, and select a date, the datepicker is automatically closed. This triggers an blur, which causes the `onFocusOut()` to be called. I think the field doesn't have a value at that point, as the 'required' error is triggered. This results in valid usage and date selection, with an invalid visible 'required' error mesage.

robinmolen: Moving the `setIsOpen(false);` (from https://github.com/open-formulieren/formio-renderer/blob/e6c4d8f87cdb2d9fe4b33cd2fe84851790c58fec/src/components/forms/DateField/DatePicker/DatePicker.tsx#L182) to outside the `flushSync` seems to solve this issue. Because the date picker is then closed, after the value is set, ensuring that the `onBlur` validation is triggered when the selected date is set.

(This might also be an issue with the datetime implementation)

src/components/forms/DateField/DatePicker/DateField-DatePicker.stories.tsx
```diff
@@ -418,3 +418,34 @@ export const ReadOnlyDatePicker: Story = {
     expect(trigger).toHaveAttribute('tabIndex', '-1');
   },
 };
+
+export const ValidateOnBlur: Story = {
```
robinmolen, line 422: Regarding the issue with premature validation:
Could you add happy flow storybook or unit tests to ensure that the validation happens correctly when using the datepicker menu to select a date, and when manually typing a date?

And could you do this for the DateTimeField as well?

src/components/forms/DatePicker.tsx
robinmolen: The validation seems also premature when you type a value in the input.

To reproduce:
1. Go to the new `validation for multiple date components` test
2. Enter a valid date `4/21/2026`
3. Click outside the date field boundary
4. Witness the "invalid date format" error message

When you click the input again, and immediately blur again, the error message is cleared. I guess this also has to do with validation being done before the actual value is set.

Again, only tested on the date component. Might also be a problem for datetime

Edit: Im noticing that the `DatePickerRoot.onFocusOut` is performed before the DatePickerField `Textbox` onBlur (which sets the value). The validation is indeed done before the field value is set.

As `onFocusOut` is performed after `onBlur`, this wasn't an issue before.

robinmolen (dismissed):
There are some premature validation issues, resulting in errors being raised when they shouldn't. Short version: the `DatePickerRoot` trigger to run field validation now happens before the field values are set.

viktorvanwijk:
> There are some premature validation issues, resulting in errors being raised when they shouldn't. Short version: the `DatePickerRoot` trigger to run field validation now happens before the field values are set.

Damn, thought it was fixed but clearly didn't test well enough. Thanks for the review :)

src/components/forms/DatePicker.tsx
viktorvanwijk: > The blur validation trigger seems to be fired prematurely when using the datepicker menu.
> 
> When you open the datepicker menu using the trigger, and select a date, the datepicker is automatically closed. This triggers an blur, which causes the `onFocusOut()` to be called. I think the field doesn't have a value at that point, as the 'required' error is triggered. This results in valid usage and date selection, with an invalid visible 'required' error mesage.

This seems to be super inconsistent. When I go to the new `Validation For Multiple Date Components` story, manually select a date in the date picker of the required field, and press submit, no errors are shown.

If there already was an error, selecting a date through the date picker will not clear it, though.

src/components/forms/DatePicker.tsx
viktorvanwijk: I have switched to the 'focusout' event, so it is accurate now :)

src/components/forms/DateField/DatePicker/DateField-DatePicker.stories.tsx
viktorvanwijk: I have added all the tests with a big disclaimer that is it really difficult to reproduce these issues in the tests. Some of them already succeeded in the current configuration, whereas performing the exact same steps manually gives a different result :upside_down_face:

src/components/forms/DatePicker.tsx
viktorvanwijk: Should be good now hopefully :crossed_fingers:

sergei-maertens (approved):

sergei-maertens:
Backported to stable/1.1.x in 47cc5438019e26e0422aaadc928b1137aa52a27a

PR #318, opened 2026-04-23
FIx the anchor target attribute being stripped

### LINKED ISSUE #6190: "target" missing in links in content component (not opening new tab)

3.4.x

### Customer reference

Leiden 51

while the link does get a `rel="noopener noreferrer"` when selecting `open in a new tab`:

Image

The `target="_blank"` is missing:

Image

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

**Tasks**

- [x] Fix in renderer (& backport)
- [x] Update SDK
    - [x] main
    - [x] stable/3.5.x
    - [x] stable/3.4.x

sergei-maertens:
Partly closes open-formulieren/open-forms#6190

### DIFF
--- a/src/registry/content/index.tsx
+++ b/src/registry/content/index.tsx
@@ -2,10 +2,18 @@ import type {ContentComponentSchema} from '@open-formulieren/types';
 import {HTMLContent} from '@utrecht/component-library-react';
 import {clsx} from 'clsx';
 import DOMPurify from 'dompurify';
+import type {Config} from 'dompurify';
 
 import type {RegistryEntry} from '@/registry/types';
 
 import './Content.scss';
+
+const DOM_PURIFY_CFG: Config = {
+  ADD_ATTR: (attributeName: string, tagName: string) => {
+    if (tagName === 'a' && attributeName === 'target') return true;
+    return false;
+  },
+};
 
 export interface FormioContentProps {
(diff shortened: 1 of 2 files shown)


src/registry/content/index.tsx
```diff
@@ -2,11 +2,19 @@ import type {ContentComponentSchema} from '@open-formulieren/types';
 import {HTMLContent} from '@utrecht/component-library-react';
 import {clsx} from 'clsx';
 import DOMPurify from 'dompurify';
+import type {Config} from 'dompurify';
 
 import type {RegistryEntry} from '@/registry/types';
 
 import './Content.scss';
 
+const DOM_PURIFY_CFG: Config = {
```
SonnyBA, line 11: We could move this configuration to a more generic file (`utils.js`?) if other components need sanitization too.

SonnyBA (approved):

src/registry/content/index.tsx
sergei-maertens: I think I'd rather have a dedicated `WYSIWYGContent` component instead that encapsulates this - we currently still have a mess in the backend with multiple editors anyway.

sergei-maertens:
Backported to stable/1.1.x in cf2151d9071470dff200e15e5f8c1b2f4983d3a0

PR #320, opened 2026-04-23
Support React 19

### LINKED ISSUE #319: Support React 19
It has been out for a while and some of our dependencies newer versions have dropped support for React 18.

For the time being we need to support both React 18 and 19, until we can confirm that the SDK has been upgraded to React 19 too.

sergei-maertens:
Closes #319

This build - with React 18 and 19 support in the peer dependencies - was tested against the SDK stable/3.5.x branch with a local:

```bash
npm run build
npm pack
# in SDK
npm i --no-save /path/to/formio-renderer/open-formulieren-formio-renderer-1.5.2.tgz
```

Tests and type checker are passing in the SDK.

I think that once this is merged, we keep the React 18 peer dependency until 3.5.x is end-of life so that can avoid stable/1.5.x+ branches as long as possible, and patches can be applied through 1.6.x etc. But, it will unblock is in upgrading the SDK to React 19 for 4.0+.

This means that at the library level, we cannot yet use the [React 19 features](https://react.dev/blog/2024/12/05/react-19#whats-new-in-react-19) though. Once we do, that means renderer 2.0 with breaking changes.


src/components/FormioForm.spec.tsx
```diff
@@ -660,7 +660,7 @@ describe('onChange prop', () => {
 
     // we expect both the user input change and the calculated clearOnHide change
     expect(onChange).toHaveBeenCalledWith({foo: 'hide', email: 'info@example.com'});
-    expect(onChange).toHaveBeenLastCalledWith({foo: 'hide'});
+    await expect.poll(() => onChange).toHaveBeenLastCalledWith({foo: 'hide'});
```
sem-maykin, line 663: What's the reason behind this change here?

sem-maykin:
Checked if the libraries support both React 18 and 19, which they do. So All clear as far as I can see. One comment on a part I don't understand yet :)

src/components/FormioForm.spec.tsx
sergei-maertens: Race condition in the callback scheduling/testing - we don't really have tight control about when React decides to call the callback inside `useEffect` (or other callback mechanisms, for that matter). I think this went right by coincidence before, or React 19's concurrency features changed an implementation detail, but essentially by the time the test code reaches this point, React may still be executing the callback.

So, the update to `expect.poll` makes the assertion retry automatically until it succeeds, which matches our behaviour that "we expect this mock to have been called with the provided arguments at the end", and it's okay if that takes a couple of ms instead of being instant.

robinmolen (approved):
Looks good 👍

src/components/FormioForm.spec.tsx
sem-maykin: I'm not very experienced with React (developed with Angular/Svelte). 

But in general, I'd say when a poll() is needed, it is showing that there is an issue in event calls. In this case when `onChange` is being triggered.

I looked a bit and if I understand correctly, `onChangeRef` is also called when `updatedValues` may be different from `values`.

PR #321, opened 2026-04-24
Support @utrecht/component-library-react 14

sergei-maertens:
It's a major version because the vega-visualization package was removed, as it was the main cause for peer dependency issues. We don't use that package/component, so it does not affect us, and upgrading will make our dependency resolution chain simpler.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "@utrecht/button-react": "^3.0.2",
     "@utrecht/calendar-react": "^1.1.2",
     "@utrecht/checkbox-react": "^1.0.13",
-    "@utrecht/component-library-react": "^13.1.0",
+    "@utrecht/component-library-react": "^14.0.0 || ^13.1.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -48,7 +48,7 @@
     "@utrecht/button-react": "^3.0.2",
     "@utrecht/calendar-react": "^1.1.2",
     "@utrecht/checkbox-react": "^1.0.13",
-    "@utrecht/component-library-react": "^13.1.0",
+    "@utrecht/component-library-react": "^14.0.0 || ^13.1.0",
```
viktorvanwijk, line 51: Switching to 14 fully breaks stuff?

viktorvanwijk (approved):

package.json
sergei-maertens: It shouldn't, but I hope this will make future backports less of a nightmare!

PR #323, opened 2026-04-24
Bandaid fix for odd conditional.eq comparison with file components

### LINKED ISSUE #6181: Use correct comparison value types in `component.conditional` when comparing against file components

3.4.8, most likely current & older too

### Customer reference

Taiga Leiden 51

When conditionally displaying a component only if one or more files are added to a file component, this conditional display doesn't work anymore in the new renderer while it did in the old one.

The condition is defined as:

```json
"conditional": {
  "show": false,
  "when": "bestandsupload",
  "eq": ""
}
```

1. Add a form with two fields:
2. File component
3. Content component & on the "Advanced" tab configure a simple conditional:
  * should display: False
  * when: file component from 2.
  * has the value: (leave blank)
4. Save the form
5. Start a form submission
6. Upload a file
7. Observe that the content component is not displayed

The content component is displayed as soon as one file is uploaded.

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
Closes open-formulieren/open-forms#6181

The builder is misconfigured and allows string values to be provided as comparison value, and the proper fix for it takes a lot more work so we apply a workaround to match the legacy renderer behaviour.

### DIFF
--- a/src/registry/file/conditional.spec.ts
+++ b/src/registry/file/conditional.spec.ts
@@ -1,0 +1,47 @@
+import type {FileComponentSchema, JSONValue} from '@open-formulieren/types';
+import type {FileUploadData} from '@open-formulieren/types/dist/components/file';
+import {expect, test} from 'vitest';
+
+import testConditional from './conditional';
+import {buildFile} from './test-utils';
+
+test.each([
+  // file components are weird and the formio-builder currently emits strings as
+  // comparison value
+  ['', [] satisfies FileUploadData[], true],
+  [
+    '',
+    [
+      buildFile({name: 'foo.png', size: 123, type: 'image/png', state: undefined}),
+    ] satisfies FileUploadData[],
+    false,
+  ],
+  ['garbage data', [] satisfies FileUploadData[], false],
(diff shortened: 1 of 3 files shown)


src/registry/file/conditional.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+  ['garbage data', [] satisfies FileUploadData[], false],
+  [
+    'garbage data',
+    [
+      buildFile({name: 'foo.png', size: 123, type: 'image/png', state: undefined}),
+    ] satisfies FileUploadData[],
+    false,
+  ],
+])(
+  'Selectboxes testConditional compares against checked/unchecked state of value',
```
viktorvanwijk:
```suggestion
  'File testConditional compares against checked/unchecked state of value',
```
;)

viktorvanwijk (approved):

sergei-maertens:
Backported to stable/1.1.x in 97b96811abc3bdf6ec1cffa73dc249249fa97763

PR #324, opened 2026-04-24
Fix addressNL crash when displayed conditionally

sergei-maertens:
Test & fix speak for themselves I'd say.


sergei-maertens:
Backported to stable/1.1.x in cdde266c9a96e04a142f154c8c7145432addb570

PR #328, opened 2026-05-28
Fix interaction bugs in `addressNL` component

### LINKED ISSUE #256: AddressNL: interaction bugs noticed during 3.4.0 testing
See form export that I used for testing: [addressnl-nieuwe-renderer.zip](https://github.com/user-attachments/files/24215735/addressnl-nieuwe-renderer.zip)

* When a postcode custom pattern is specified with custom error message, the custom error message is not displayed during client side validation
* When a custom city pattern is specified, this may not ever run during client side activation - it can't be properly tested/verified because the input field is `disabled`
* When entering a postcode and house number (`1111 ZZ`, `99999`) that doesn't resolve to any address, the city and street are required and invalid, blocking the step submit button. When the inputs are corrected, the step submit button is still disabled where you'd expect it to be enabled again. I suspect this has to do with the field not being valid and the disabled nature making it impossible to retrigger validation.

sergei-maertens:
Closes #256

- [x] Improved validation error handling when custom postcode validation error messages are configured
- [x] Ensure validation errors on the (read-only) city field are displayed if there are any
- [x] Display a useful message if no addresses could be resolved from the postcode / house number inputs (invalid address)
- [x] Fix validation state properly updating when error are corrected

### DIFF
--- a/src/registry/addressNL/subFields.tsx
+++ b/src/registry/addressNL/subFields.tsx
@@ -37,6 +37,7 @@ interface PostCodeFieldProps {
 export const PostCodeField: React.FC<PostCodeFieldProps> = ({namePrefix, isRequired}) => (
   <TextField
     name={`${namePrefix}.postcode`}
+    nameForValidate={namePrefix.split('.')[0]}
     label={<FormattedMessage {...FIELD_LABELS.postcode} />}
     placeholder="1234 AB"
     isRequired={isRequired}
@@ -51,6 +52,7 @@ interface HouseNumberFieldProps {
 export const HouseNumberField: React.FC<HouseNumberFieldProps> = ({namePrefix, isRequired}) => (
   <TextField
     name={`${namePrefix}.houseNumber`}
+    nameForValidate={namePrefix.split('.')[0]}
     label={<FormattedMessage {...FIELD_LABELS.houseNumber} />}
     inputMode="numeric"
     placeholder="123"
@@ -65,6 +67,7 @@ interface HouseLetterProps {
 export const HouseLetter: React.FC<HouseLetterProps> = ({namePrefix}) => (
   <TextField
     name={`${namePrefix}.houseLetter`}
+    nameForValidate={namePrefix.split('.')[0]}
     label={<FormattedMessage {...FIELD_LABELS.houseLetter} />}
   />
 );
@@ -76,6 +79,7 @@ interface HouseNumberAdditionProps {
 export const HouseNumberAddition: React.FC<HouseNumberAdditionProps> = ({namePrefix}) => (
   <TextField
     name={`${namePrefix}.houseNumberAddition`}
+    nameForValidate={namePrefix.split('.')[0]}
     label={<FormattedMessage {...FIELD_LABELS.houseNumberAddition} />}
   />
 );
@@ -89,6 +93,7 @@ interface StreetNameProps {
 export const StreetName: React.FC<StreetNameProps> = ({namePrefix, isRequired, isReadOnly}) => (
   <TextField
     name={`${namePrefix}.streetName`}
+    nameForValidate={namePrefix.split('.')[0]}
     label={<FormattedMessage {...FIELD_LABELS.streetName} />}
     isRequired={isRequired}
     isReadOnly={isReadOnly}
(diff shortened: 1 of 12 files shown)


src/registry/addressNL/validationSchema.ts
```diff
@@ -34,6 +34,11 @@ const HOUSE_NUMBER_ADDITION_INVALID_MESSAGE = defineMessage({
   defaultMessage: 'House number addition must be up to four letters and digits.',
 });
 
+const INVALID_AUTOFILL_ADDRESS_MESSAGE = defineMessage({
+  description: 'Validation error for empty address derivation result.',
+  defaultMessage: "The entered postcode and number don't point to a known address.",
```
robinmolen:
```suggestion
  defaultMessage: "The entered postcode and number don't result into a known address.",
```

src/registry/addressNL/validation.stories.ts
```diff
@@ -265,3 +265,47 @@ export const AutofillDoesNotResolve: Story = {
...
+    });
+
+    await step('Clear original postcode field', async () => {
+      const postcodeField = canvas.getByLabelText('Postcode');
+      await userEvent.clear(postcodeField);
+      await userEvent.keyboard('{Tab}');
+    });
+
+    await step('Submit', async () => {
+      await waitFor(() => expect(submitButton).not.toHaveAttribute('data-is-valid', 'false'));
```
robinmolen: ~`data-is-valid` should be `true` right? As the error messages are cleared and the form is now valid?~

Why not explicitly expect the `data-is-valid` to have the value `true`?:
```suggestion
      await waitFor(() => expect(submitButton).toHaveAttribute('data-is-valid', 'true'));
```

robinmolen (changes requested):
Two small suggestions. Otherwise, looks good

src/registry/addressNL/validation.stories.ts
sergei-maertens: Because I first had a test based on `aria-disabled` :-) but that broke a lot of other tests, so reworked it to a data-attribute. Will update, this is cleaner!

src/registry/addressNL/subFields.tsx
```diff
@@ -37,6 +37,7 @@ interface PostCodeFieldProps {
 export const PostCodeField: React.FC<PostCodeFieldProps> = ({namePrefix, isRequired}) => (
   <TextField
     name={`${namePrefix}.postcode`}
+    nameForValidate={namePrefix.split('.')[0]}
```
robinmolen, line 40: This results in the top-most component name, right? So for nested addressNL components this will result in the name of the parent component. And in cases of deep nested components (addressNL in editgrid in edigrid), it will result in the top-most editgrid name.

Is that what we want? Should it not be `namePrefix.split('.').pop()` to get the name of the addressNL component itself? (This applies to all AddressNL sub-fields)

robinmolen (approved):

src/registry/addressNL/subFields.tsx
sergei-maertens: if you have an addressNL component with key `foo.bar.addressNL`, this:

* results in a data structure `{"foo": {"bar": {"addressNL": {"postcode": "..."}}}}`
* consequently, if there are no validation errors anymore (after addressing that, this leads to a error structure like: `{"foo": {"bar": {"addressNL": {}}}}`
* to properly clear the errors, we must validate at the outermost layer so that we get `errors = {}`, any intermediate will be seen by Formik as "not valid"

So this effectively means that we validate all the data at path `"foo"`. Meaning, if there's another component that has key `foo.baz`, it will also be validated.

Your suggestion would validate `addressNL`, which is not an existing path in the formik values/errors (it points to a top-level key `addressNL`)

So unfortunately, this is the compromise to make.

Note that validation errors are only displayed when a field has been touched, and we don't see these dotted paths *too* often so I think it will be okay.

sergei-maertens: Via Slack another bug became apparent: #329 tracking that separately

sergei-maertens:
Backport to 3.5.x will be done with a regular dependency bump, but adding the label to stay aware of it.

PR #330, opened 2026-06-02
Fix validation state with "nested data" through periods in component keys blocking submission

### LINKED ISSUE #329: Validation errors/state don't fully clear for form values with nested structures
Field validation is hooked up to the name of field, e.g. `foo.bar`, which only validates the leaf node. However, this can result in left-over non-empty `errors` state in Formik that blocks submission because `isValid` is falsy.

Steps to reproduce:

1. Have a submit button that's disabled if `useFormikContext().isValid` is `false`
2. Define a formio configuration with a textfield component with key `foo.bar` and a validation rule (e.g. max length 3).
3. Type 4 characters and blur the field to trigger validation. You should now see a validation error.
4. The submit button is now disabled.
5. Remove a character to clear the validation error and blur the field again.
6. The error is now gone
7. The submit button is still disabled, while it should have been cleared.

I see two options to fix this:

* always validate the top-level node for fields (approach taken with addressNL)
* don't use Formik's `isValid`, but instead calculate it ourselves if we detect that the `errors` structure only contains empty nodes.

sergei-maertens:
Closes #329

### DIFF
--- a/src/components/FormioForm.spec.tsx
+++ b/src/components/FormioForm.spec.tsx
@@ -966,4 +966,55 @@ describe('Regressions', () => {
     await screen.getByRole('button', {name: 'Submit'}).click();
     await expect.poll(() => onSubmit).toHaveBeenCalledWith({container: {field: '123'}});
   });
+
+  test('Correctly reports validity state with periods in component keys inside editgrids', async () => {
+    const onSubmit = vi.fn();
+    const screen = await render(
+      <Form
+        components={[
+          {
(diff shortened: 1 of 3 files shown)


robinmolen (approved):
Nice! I expected it to be much more of a pain 😅

Lets merge it!

sergei-maertens:
Backport is scoped to SDK 3.5.+, so we can use the regular dependency update and don't bother with 1.1.x (used in OF 3.4).

PR #331, opened 2026-06-02
Upgrade dependencies

sergei-maertens:
None of the updated packages are known to be in a supply chain attack currently/recently :crossed_fingers:


vaszig (approved):
I accepted the UI tests too but I see a lot of times the map to be inconsistent.

PR #334, opened 2026-06-04
[#6315] Add modal sheet variant

### LINKED ISSUE #6315: Add support for configurable bottom-sheet tool tips
- [x] Implement bottom-sheet tool tip following the design in the front end  https://github.com/open-formulieren/formio-renderer/pull/334
- [x] Implement tool tip per field following the design in the front end https://github.com/open-formulieren/formio-renderer/pull/335
  - [x] Update field component schemas to include newly configurable tool tip related fields https://github.com/open-formulieren/types/pull/95
- [x] Allow multiple tool tips to be configured per field through the admin form builder https://github.com/open-formulieren/formio-builder/pull/300

**Additional notes for the form admin**
-  Allow bullet points & bold styles for WYSIWYG editor
- Configurable fields for the tool tips:
  - Title
  - Content

SonnyBA:
Closes the first bullet point in https://github.com/open-formulieren/open-forms/issues/6315.

### DIFF
--- a/src/components/modal/Modal.scss
+++ b/src/components/modal/Modal.scss
@@ -7,6 +7,15 @@ body:has(.openforms-modal[open]) {
   overflow: hidden;
 }
 
+@keyframes slide-in {
+  0% {
+    transform: translateY(100%);
+  }
+  100% {
+    transform: translateY(0%);
+  }
+}
+
 .openforms-modal {
   --_of-modal-padding-block-start: var(--of-modal-padding-block-start, 40px);
   --_of-modal-padding-block-end: var(--of-modal-padding-block-end, 40px);
@@ -39,6 +48,25 @@ body:has(.openforms-modal[open]) {
     0 6px 20px rgba(0, 0, 0, 0.2)
   );
   box-sizing: border-box;
+
+  @include bem.modifier('sheet') {
+    max-block-size: var(--of-modal-sheet-max-block-size, 75dvh);
+    inline-size: var(--of-modal-sheet-inline-size, 100dvw);
+    max-inline-size: var(--of-modal-sheet-max-inline-size, 100vw);
+
+    border-radius: var(--of-modal-sheet-border-radius, 5px 5px 0 0);
+
+    margin-block-end: 0;
+
+    animation-duration: 0.2s;
+    animation-name: slide-in;
+
+    @include responsive.desktop {
+      margin-block-end: revert;
+
+      animation: none;
+    }
+  }
 
   &[open] {
     display: flex;
@@ -132,5 +160,11 @@ body:has(.openforms-modal[open]) {
     padding-block-start: var(--_of-modal-body-bleed-margin);
     padding-inline-end: var(--_of-modal-body-bleed-margin);
     padding-inline-start: var(--_of-modal-body-bleed-margin);
+  }
+}
+
+.openforms-theme {
+  @include responsive.desktop {
+    --of-modal-sheet-inline-size: 60dvw;
   }
 }
--- a/src/components/modal/Modal.stories.tsx
+++ b/src/components/modal/Modal.stories.tsx
@@ -58,6 +58,8 @@ type Story = StoryObj<typeof Modal>;
 
 export const Default: Story = {};
 
+export const Sheet: Story = {args: {variant: 'sheet'}};
+
 export const CloseModal: Story = {
   decorators: [withTriggerDecorator],
   args: {
@@ -103,9 +105,24 @@ export const CloseModal: Story = {
   },
 };
 
+export const CloseSheet: Story = {
+  ...CloseModal,
+  args: {
+    ...CloseModal.args,
+    variant: 'sheet',
+  },
+};
+
 export const WithPortal: Story = {
   args: {
     noPortal: false,
+  },
+};
+
+export const WithPortalSheet: Story = {
+  args: {
+    noPortal: false,
+    variant: 'sheet',
   },
 };
 
@@ -159,5 +176,13 @@ export const OutlinesVisible: Story = {
 
     const checkbox = await canvas.findByRole('checkbox', {name: /On focus/});
     checkbox.focus();
+  },
+};
+
+export const OutlinesVisibleSheet: Story = {
+  ...OutlinesVisible,
+  args: {
+    ...OutlinesVisible.args,
+    variant: 'sheet',
   },
 };
--- a/src/components/modal/Modal.tsx
+++ b/src/components/modal/Modal.tsx
@@ -1,4 +1,5 @@
 import {Heading2, Icon as UtrechtIcon} from '@utrecht/component-library-react';
+import {clsx} from 'clsx';
 import {useEffect, useId, useRef} from 'react';
 import ReactDOM from 'react-dom';
 import {useIntl} from 'react-intl';
@@ -9,6 +10,8 @@ import Icon from '@/components/icons';
 import './Modal.scss';
 import {useModalContext} from './hooks';
 
+export type ModalVariant = 'modal' | 'sheet';
+
 export interface ModalProps {
   isOpen?: boolean;
   title: React.ReactNode;
@@ -16,6 +19,7 @@ export interface ModalProps {
   closeModal: () => void;
   children: React.ReactNode;
   noPortal?: boolean;
+  variant?: ModalVariant;
 }
 
 const Modal: React.FC<ModalProps> = ({
@@ -24,6 +28,7 @@ const Modal: React.FC<ModalProps> = ({
   titleId,
   closeModal,
   noPortal = false,
+  variant = 'modal',
   children,
 }) => {
   const id = useId();
@@ -51,7 +56,7 @@ const Modal: React.FC<ModalProps> = ({
 
   const dialogUI = (
     <dialog
-      className="openforms-modal"
+      className={clsx('openforms-modal', {'openforms-modal--sheet': variant === 'sheet'})}
       ref={modalRef}
       aria-labelledby={titleId}
       onClose={closeModal}


src/components/modal/Modal.stories.tsx
```diff
@@ -101,16 +108,27 @@ export const CloseModal: Story = {
       });
     });
   },
+});
+
+export const CloseModal: Story = createCloseStory('modal');
+export const CloseSheet: Story = createCloseStory('sheet');
```
sergei-maertens: Can you avoid the factory function and stick to plain story definitions please?

If you want to re-use stories, you can use the prop spreading:

```tsx
export const CloseSheet: Story = {
    ...CloseModal,
    args: {
        variant: ...
    },
};
```

src/components/modal/Modal.tsx
```diff
@@ -7,23 +7,33 @@ import {SubtleButton} from '@/components/Button';
 import Icon from '@/components/icons';
 
 import './Modal.scss';
+import './Sheet.scss';
```
sergei-maertens: it's still the same component - we stick to a single SCSS file per component as entrypoint. You can move the CSS into `Modal.scss`

src/components/modal/Modal.tsx
```diff
@@ -7,23 +7,33 @@ import {SubtleButton} from '@/components/Button';
...
   closeModal: () => void;
   children: React.ReactNode;
   noPortal?: boolean;
+  variant?: ModalVariant;
 }
 
+const modalClassMap: Record<ModalVariant, string> = {
+  modal: 'openforms-modal',
+  sheet: 'openforms-sheet',
+};
```
sergei-maertens: can't this be a modifier instead of a completely different class? `openforms-modal openforms-modal--sheet` properly indicates it's a layout variant

sergei-maertens (changes requested):

src/components/modal/Modal.scss
```diff
@@ -40,6 +51,21 @@ body:has(.openforms-modal[open]) {
   );
   box-sizing: border-box;
 
+  &--sheet {
```
sergei-maertens: please use the available BEM helpers. Your code produces a selector like `.openforms-modal--sheet`, the BEM helper produces a seletor `.openforms-modal.openforms-modal--sheet` which is more strict.

The former allows invalid class names on an element with only `openforms-modal--sheet`, while the strict variant requires both class names. You always need a base if you use a modifier: either the block or element (`.foo` or `.foo__bar`) to apply the modifiers to.


```suggestion
  @include bem.modifier('sheet') {
```

src/components/modal/Modal.scss
```diff
@@ -40,6 +51,21 @@ body:has(.openforms-modal[open]) {
...
+    max-block-size: 75dvh;
+    inline-size: 100dvw;
```
sergei-maertens: Let's make those design tokens since we already got questions about these values being hardcoded...


```suggestion
    max-block-size: var(--of-modal-sheet-max-block-size, 75dvh);
    inline-size: var(--of-modal-sheet-inline-size, 100dvw);
```

src/components/modal/Modal.scss
```diff
@@ -40,6 +51,21 @@ body:has(.openforms-modal[open]) {
...
+
+    // override browser defaults
+    max-width: 100vw;
```
sergei-maertens: it's not a browser default, it's from the `.openforms-modal` base class I believe


```suggestion
    // override browser defaults
    max-inline-size: 100vw;
```

src/components/modal/Modal.scss
```diff
@@ -40,6 +51,21 @@ body:has(.openforms-modal[open]) {
...
+
+    margin-bottom: 0;
```
sergei-maertens:
```suggestion
    margin-block-end: 0;
```
favour logical properties over top/bottom/left/right, especially with right-to-left languages in the future you'll be able to write agnostic CSS.

Additionally, this pins the modal to the bottom on desktop, which is not according to the design where it's centered on the page. Only on mobile should it be pinned to the bottom.

src/components/modal/Modal.scss
```diff
@@ -15,6 +24,8 @@ body:has(.openforms-modal[open]) {
 
   --_of-modal-body-bleed-margin: var(--of-modal-body-bleed-margin, 2px);
 
+  --_of-modal-border-radius: var(--of-sheet-border-radius, 5px 5px 0 0);
```
sergei-maertens: I'm not sure why you're defining a private/locally scoped CSS variable here?

Additionally, design tokens must follow the format `--of-{component}-{variant/aspect/...}-{property}`, so the right design token name would be `--of-modal-sheet-border-radius`

src/components/modal/Modal.scss
```diff
@@ -40,6 +51,21 @@ body:has(.openforms-modal[open]) {
...
+  &--sheet {
+    max-block-size: 75dvh;
+    inline-size: 100dvw;
+
+    // override browser defaults
+    max-width: 100vw;
+
+    margin-bottom: 0;
+
+    border-radius: var(--_of-sheet-border-radius);
```
sergei-maertens:
```suggestion
    border-radius: var(--of-modal-sheet-border-radius, 5px 5px 0 0);
```

src/components/modal/Modal.scss
```diff
@@ -54,6 +80,13 @@ body:has(.openforms-modal[open]) {
     inline-size: 50dvw;
     min-block-size: 30dvh;
     max-block-size: 50dvh;
+
+    &--sheet {
+      inline-size: 60dvw;
+      max-block-size: 75dvh;
+
+      animation: none;
+    }
```
sergei-maertens: I'd inlcude the media query in the upper `include bem.modifier('sheet') {...}` block above so that styles for a single modifier are contained. I'm also afraid that in this current version of the BEM stuff, you can't yet use `@include ...` inside a media query, so that solves two "problems".

The hardcoded inline-size also doesn't seem practical, but we also should not define design tokens anymore for responsiveness.

Can you try with an Open Forms theme specific code block like this to see if this works?

```css
.openforms-theme {
  @include responsive.desktop {
    --of-modal-sheet-inline-size: 60dvw;
  }
}
```

If that works, it'd be the official way for custom themes to manage the responsive sizing.

src/components/modal/Modal.tsx
```diff
@@ -51,7 +56,7 @@ const Modal: React.FC<ModalProps> = ({
 
   const dialogUI = (
     <dialog
-      className="openforms-modal"
+      className={clsx('openforms-modal', variant === 'sheet' && 'openforms-modal--sheet')}
```
sergei-maertens:
```suggestion
      className={clsx('openforms-modal', {'openforms-modal--sheet': variant === 'sheet'})}
```
clsx has this boolean-switch mechanism built in :)

src/components/modal/Modal.tsx
```diff
@@ -65,12 +70,26 @@ const Modal: React.FC<ModalProps> = ({
       // eslint-disable-next-line react/no-unknown-property
       closedby="any"
     >
-      <header className="openforms-modal__header">
-        <Heading2 id={titleId} className="openforms-modal__title">
+      <header
+        className={clsx(
+          'openforms-modal__header',
+          variant === 'sheet' && 'openforms-modal__header--sheet'
```
sergei-maertens: why is this necesary?

You can already target `.openforms-modal .openforms-modal--sheet openforms-modal__header` in CSS if necessary - the elements/structure stay the same and the block itself can have a variant/modifier.

src/components/modal/Modal.tsx
```diff
@@ -65,12 +70,26 @@ const Modal: React.FC<ModalProps> = ({
...
+        )}
+      >
+        <Heading2
+          id={titleId}
+          className={clsx(
+            'openforms-modal__title',
+            variant === 'sheet' && 'openforms-modal__title--header'
```
sergei-maertens: what is a `header` modifier for?

src/components/modal/Modal.tsx
```diff
@@ -65,12 +70,26 @@ const Modal: React.FC<ModalProps> = ({
...
+          )}
+        >
           {title}
         </Heading2>
         <SubtleButton
-          className="openforms-modal__close"
+          className={clsx(
+            'openforms-modal__close',
+            variant === 'sheet' && 'openforms-modal__close--sheet'
```
sergei-maertens: see above - I'm not seeing any CSS overrides for these variants either.

src/components/modal/Modal.tsx
```diff
@@ -84,7 +103,14 @@ const Modal: React.FC<ModalProps> = ({
         </SubtleButton>
       </header>
 
-      <div className="openforms-modal__body">{children}</div>
+      <div
+        className={clsx(
+          'openforms-modal__body',
+          variant === 'sheet' && 'openforms-modal__body--sheet'
```
sergei-maertens: see above

sergei-maertens (changes requested):

src/components/modal/Modal.scss
SonnyBA: Should these new variables be added to the design tokens repository ([to the modal file](https://github.com/open-formulieren/design-tokens/blob/main/src/components/modal.tokens.json#L3))?

sergei-maertens: Preferably yes, because then they show up in documentation. Though technically it'll also work fine with the defaults that are provided :)

SonnyBA: See https://github.com/open-formulieren/design-tokens/pull/69 for the PR in the design-tokens repository.

sergei-maertens (approved):

PR #335, opened 2026-06-09
[#6315] Add FAQ tooltip

### LINKED ISSUE #6315: Add support for configurable bottom-sheet tool tips

SonnyBA:
Closes the following bullet point, [from this ticket](https://github.com/open-formulieren/open-forms/issues/6315#issue-4523952977):

> Implement tool tip per field following the design in the front end

This branch was branched off from `open-formulieren/feature/6315-modal-sheet` as it depends on the updated Modal component.

TODO:

- [x] Update the pinned types version (to use the `FAQItem` types)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "EUPL-1.2",
       "dependencies": {
         "@babel/runtime": "^7.29.2",
-        "@open-formulieren/types": "^1.3.0",
+        "@open-formulieren/types": "^1.4.2",
         "clsx": "^2.1.0",
         "dompurify": "^3.4.11",
         "formik": "^2.4.5",
@@ -2214,9 +2214,9 @@
       }
     },
     "node_modules/@open-formulieren/types": {
-      "version": "1.3.0",
-      "resolved": "https://registry.npmjs.org/@open-formulieren/types/-/types-1.3.0.tgz",
-      "integrity": "sha512-fSzo0glfvIJy0e23tjNst/eiPm+TgY/mD1jCsS9QpggdOQQe2MdEaj5h94yJTEZcgtlXyv9sOjEOVoKT4XKtmg==",
+      "version": "1.4.2",
+      "resolved": "https://registry.npmjs.org/@open-formulieren/types/-/types-1.4.2.tgz",
+      "integrity": "sha512-WLBk3FSEMsJFIHF/MnD7gBMk4Oi5ZdkY8LWAhYgjfSpJXYv94wIeoOtQqJ3UT/9kXzpoxi+VFEC4n9cikT/DVg==",
       "license": "EUPL-1.2"
     },
     "node_modules/@oxc-parser/binding-android-arm-eabi": {
@@ -2346,9 +2346,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2366,9 +2363,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2386,9 +2380,6 @@
         "ppc64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2406,9 +2397,6 @@
         "riscv64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2426,9 +2414,6 @@
         "riscv64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2446,9 +2431,6 @@
         "s390x"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2466,9 +2448,6 @@
         "x64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2486,9 +2465,6 @@
         "x64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2720,9 +2696,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2737,9 +2710,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2754,9 +2724,6 @@
         "ppc64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2771,9 +2738,6 @@
         "riscv64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2788,9 +2752,6 @@
         "riscv64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2805,9 +2766,6 @@
         "s390x"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2822,9 +2780,6 @@
         "x64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2839,9 +2794,6 @@
         "x64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
--- a/package.json
+++ b/package.json
@@ -177,7 +177,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.29.2",
-    "@open-formulieren/types": "^1.3.0",
+    "@open-formulieren/types": "^1.4.2",
     "clsx": "^2.1.0",
     "dompurify": "^3.4.11",
     "formik": "^2.4.5",
--- a/src/components/forms/Checkbox/Checkbox.stories.ts
+++ b/src/components/forms/Checkbox/Checkbox.stories.ts
@@ -52,6 +52,31 @@ export const WithTooltip: Story = {
   },
 };
 
+export const WithFAQItems: Story = {
+  args: {
+    isRequired: true,
+    faqItems: [
+      {
+        label: 'How do I fill in this field?',
+        content: 'The values required to fill out this field can be retrieved from XYZ.',
+      },
+      {
+        label: 'Is this field applicable to me?',
+        content: 'This field is applicable if you are XYZ.',
+      },
+    ],
+  },
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+    const faqLabels1 = canvas.getByRole('button', {name: 'How do I fill in this field?'});
+    const faqLabels2 = canvas.getByRole('button', {name: 'Is this field applicable to me?'});
+
+    // Label should be visible underneath the fields
+    await expect(faqLabels1).toBeVisible();
+    await expect(faqLabels2).toBeVisible();
+  },
+};
+
 export const ValidationError: Story = {
   name: 'Validation error',
   args: {
--- a/src/components/forms/Checkbox/Checkbox.tsx
+++ b/src/components/forms/Checkbox/Checkbox.tsx
@@ -1,3 +1,4 @@
+import type {FAQItem} from '@open-formulieren/types';
 import {Checkbox as UtrechtCheckbox} from '@utrecht/checkbox-react';
 import {FormFieldDescription} from '@utrecht/form-field-description-react';
 import {FormField} from '@utrecht/form-field-react';
@@ -5,6 +6,7 @@ import {clsx} from 'clsx';
 import {useField, useFormikContext} from 'formik';
 import {useId} from 'react';
 
+import FAQItems from '@/components/forms/FAQItems';
 import HelpText from '@/components/forms/HelpText';
 import {LabelContent} from '@/components/forms/Label';
 import Tooltip from '@/components/forms/Tooltip';
@@ -58,6 +60,11 @@ export interface CheckboxProps {
    * assist users in filling out the field correctly.
    */
   tooltip?: React.ReactNode;
+  /**
+   * Optional FAQ tooltips to provide additional information that is not crucial but may
+   * assist users in filling out the field correctly.
+   */
+  faqItems?: FAQItem[];
 }
 
 const Checkbox: React.FC<CheckboxProps> = ({
@@ -69,6 +76,7 @@ const Checkbox: React.FC<CheckboxProps> = ({
   isReadOnly = false,
   ignoreRequired = false,
   tooltip,
+  faqItems = [],
 }) => {
   const {validateField} = useFormikContext();
   name = useFieldConfig(name);
@@ -86,6 +94,7 @@ const Checkbox: React.FC<CheckboxProps> = ({
   const descriptionId = description && !descriptionAsHelpText ? `${id}-description` : undefined;
 
   const ariaDescribedBy = [errorMessageId, descriptionId].filter(Boolean).join(' ');
+
   return (
     <FormField type="checkbox" invalid={invalid} className="utrecht-form-field--openforms">
       <UtrechtCheckbox
@@ -135,6 +144,7 @@ const Checkbox: React.FC<CheckboxProps> = ({
       )}
 
       {touched && errorMessageId && <ValidationErrors error={error} id={errorMessageId} />}
+      <FAQItems items={faqItems} />
     </FormField>
   );
 };
--- a/src/components/forms/DateField/DateField.tsx
+++ b/src/components/forms/DateField/DateField.tsx
@@ -1,7 +1,9 @@
+import type {FAQItem} from '@open-formulieren/types';
 import {FormField} from '@utrecht/form-field-react';
 import {useFormikContext} from 'formik';
 import {useId} from 'react';
 
+import FAQItems from '@/components/forms/FAQItems';
 import HelpText from '@/components/forms/HelpText';
 import ValidationErrors from '@/components/forms/ValidationErrors';
 import {useFieldConfig, useFieldError} from '@/hooks';
@@ -81,6 +83,11 @@ interface DateFieldCommonProps {
    */
   tooltip?: React.ReactNode;
   /**
+   * Optional FAQ tooltips to provide additional information that is not crucial but may
+   * assist users in filling out the field correctly.
+   */
+  faqItems?: FAQItem[];
+  /**
    * How to autocomplete the field from the browser.
    */
   autoComplete?: string;
@@ -112,6 +119,7 @@ const DateField: React.FC<DateFieldProps> = ({
   tooltip,
   autoComplete,
   isMultiValue = false,
+  faqItems = [],
   ...props
 }) => {
   const id = useId();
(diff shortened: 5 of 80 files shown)


src/components/forms/Checkbox/Checkbox.tsx
```diff
@@ -86,6 +94,10 @@ const Checkbox: React.FC<CheckboxProps> = ({
   const descriptionId = description && !descriptionAsHelpText ? `${id}-description` : undefined;
 
   const ariaDescribedBy = [errorMessageId, descriptionId].filter(Boolean).join(' ');
+  const faqElements = faqItems.map((faqItem, index) => (
+    <FAQTooltip key={index} faqItem={faqItem} />
+  ));
```
sergei-maertens: this map is repeated everywhere - wouldn't it make more sense to have a simple component that you call like this?

```tsx
<FAQItems items={faqItems} />
```
The map/loop itself would then be implement in the `FAQItems` component

sergei-maertens (changes requested):

src/components/forms/Checkbox/Checkbox.stories.ts
```diff
@@ -52,6 +52,35 @@ export const WithTooltip: Story = {
...
+      {
+        label: 'Is this field applicable to me?',
+        content: 'This field is applicable if you are XYZ.',
+      },
+    ],
+  },
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+    const faqLabels1 = canvas.getAllByText('How do I fill in this field?');
+    const faqLabels2 = canvas.getAllByText('Is this field applicable to me?');
```
sergei-maertens: queries `byRole` are better as they ensure that you have the accessibility requirements right - in this case, you'd do:

```tsx
canvas.getByRole('button', {name: 'How do I fill in this field?'});
```

src/components/forms/Checkbox/Checkbox.tsx
```diff
@@ -11,6 +12,7 @@ import Tooltip from '@/components/forms/Tooltip';
 import ValidationErrors from '@/components/forms/ValidationErrors';
 import {useFieldConfig} from '@/hooks';
 
+import FAQItems from '../FAQItems';
```
sergei-maertens:
```suggestion
import FAQItems from '@/components/forms/FAQItems';
```
please use absolute imports - typically we use relative imports only when in the same 'package`, so only `./Foo` would be allowed.

src/components/forms/DateField/DateField.tsx
```diff
@@ -158,6 +166,7 @@ const DateField: React.FC<DateFieldProps> = ({
     <FormField type="text" invalid={isInvalid} className="utrecht-form-field--openforms">
       {dateInput}
       <HelpText>{description}</HelpText>
+      <FAQItems items={faqItems} />
```
sergei-maertens: I think you want validation errors to go above faqitems - if you have 5 items, it would push the errors off the screen.

package-lock.json
```diff
@@ -2346,9 +2347,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
```
sergei-maertens, line 2351: where is this coming from? Are you perhaps using a different node version than what's in `.nvmrc`?

package.json
```diff
@@ -77,6 +77,7 @@
     "@maykinmedia/eslint-config": "^2.0.0",
     "@open-formulieren/design-tokens": "^0.66.0",
     "@open-formulieren/leaflet-tools": "^1.0.0",
+    "@open-formulieren/types": "^1.4.1",
```
sergei-maertens: why was this moved into the dev dependencies?

src/components/forms/FAQItems.tsx
sergei-maertens, line 1: I would keep all the code related to the FAQItems together, ultimately that's the "public" API for the component that you expose to other components. `FAQTooltip` can be moved into this file, and doesn't need to be exported then.

src/components/forms/FAQItems.tsx
```diff
@@ -0,0 +1,17 @@
+import type {FAQItem} from '@open-formulieren/types/dist/common';
+
+import FAQTooltip from './FAQTooltip';
+
+export interface CheckboxProps {
+  /**
+   * FAQ tooltips to provide additional information that is not crucial but may
+   * assist users in filling out the field correctly.
+   */
+  items?: FAQItem[];
```
sergei-maertens:
```suggestion
  items: FAQItem[];
```
be strict if possible - this component makes no sense if there are no item definitions - the call site can and should opt to pass a empty array as default or conditionally render: `faqItems && <FAQItems items={faqItems} />`.

That leads to a simpler implementation and less edge cases to test for this component.

src/components/forms/FAQItems.tsx
```diff
@@ -0,0 +1,17 @@
+import type {FAQItem} from '@open-formulieren/types/dist/common';
+
+import FAQTooltip from './FAQTooltip';
+
+export interface CheckboxProps {
```
sergei-maertens: `CheckboxProps`?

src/components/forms/FAQItems.tsx
```diff
@@ -0,0 +1,17 @@
+import type {FAQItem} from '@open-formulieren/types/dist/common';
```
sergei-maertens: I'd update the types to have this exported from the entrypoint: https://github.com/open-formulieren/types/blob/main/src/index.ts


```suggestion
import type {FAQItem} from '@open-formulieren/types';
```

src/components/forms/FAQItems.tsx
```diff
@@ -0,0 +1,17 @@
...
+export interface CheckboxProps {
+  /**
+   * FAQ tooltips to provide additional information that is not crucial but may
+   * assist users in filling out the field correctly.
+   */
+  items?: FAQItem[];
+}
+
+const FAQItems: React.FC<CheckboxProps> = ({items = []}) => {
+  return items.map((item, index) => <FAQTooltip key={index} faqItem={item} />);
```
sergei-maertens:
```suggestion
  return items.map(({label, content}, index) => <FAQTooltip key={index} label={label} content={content} />);
```
this is much more idiomatic react - you're passing primitives as props instead of an object, which works better with potential future memoization and is much more explicit + more ergonomic when using this component in isolation.

In its current form, you have some excessive nesting and if the `FAQItem` type gets extended with stuff that's not directly relevant for these components, you're dealing with friction and unrelated changes when another PR bumps the types version (you'll need to update stories and call sites).

src/components/forms/FAQTooltip.tsx
```diff
@@ -0,0 +1,52 @@
+import type {FAQItem} from '@open-formulieren/types/dist/common';
+import {HTMLContent, Icon as UtrechtIcon} from '@utrecht/component-library-react';
+import DOMPurify from 'dompurify';
+import type {KeyboardEvent} from 'react';
+import {useState} from 'react';
+
+import Icon from '@/components/icons';
+
+import Modal from '../modal';
```
sergei-maertens: please avoid relative imports to parent directories, it's so hard to see where the component is being imported from when reviewing a PR

src/components/forms/FAQTooltip.tsx
```diff
@@ -0,0 +1,52 @@
...
+import './FAQTooltip.scss';
+
+export interface FAQTooltipProps {
+  faqItem: FAQItem;
+}
```
sergei-maertens: as mentioned above - let's split this out and make it explicit. It's Typescript's job to make sure that our FAQItem at call sites matches the expected props :)

src/components/forms/FAQTooltip.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+const FAQTooltip: React.FC<FAQTooltipProps> = ({faqItem}) => {
```
sergei-maertens: Given the DOMPurify sanitize happening on every render (and the parent rerender triggering a re-render of all children), this is potentially expensive. Combined with making the props primitives and atomic, I would immediately opt for [`React.memo`](https://react.dev/reference/react/memo) to avoid performance issues in large forms with many faq items and/or complex content markup.

src/components/forms/FAQTooltip.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const [isOpen, setIsOpen] = useState(false);
+
+  const sanitizedContent = DOMPurify.sanitize(faqItem.content, {USE_PROFILES: {html: true}});
+  const modalSheet = (
+    <Modal
+      title={faqItem.label}
+      variant={'sheet'}
```
sergei-maertens:
```suggestion
      variant="sheet"
```

src/components/forms/FAQTooltip.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const dialogOpenKeys = ['Enter', 'Space'];
+  return (
+    <>
+      <div
+        role="button"
+        tabIndex={0}
+        className="openforms-faq-tooltip"
+        onClick={() => setIsOpen(true)}
+        onKeyDown={(event: KeyboardEvent) => dialogOpenKeys.includes(event.code) && setIsOpen(true)}
+      >
```
sergei-maertens: This really should be a proper `<button>` element instead of a div pretending to a button. Also, we really should use existing NL DS components to make this themeable.

I would however opt for a custom modifier so that custom themes can apply CSS overrides if necessary: a final  classname of `utrecht-link-button openforms-faq-item` should get us there for all the styling needs.

src/components/forms/FAQTooltip.tsx
```diff
@@ -0,0 +1,52 @@
...
+        <UtrechtIcon>
+          <Icon icon="tooltip" className="openforms-faq-tooltip__icon" />
```
sergei-maertens:
```suggestion
          <Icon icon="tooltip" />
```
I don't think the additional class name is necessary

src/components/forms/FAQTooltip.tsx
```diff
@@ -0,0 +1,52 @@
...
+        </UtrechtIcon>
+
+        <span className="openforms-faq-tooltip__title">{faqItem.label}</span>
```
sergei-maertens:
```suggestion
        <span className="openforms-faq-item__title">{faqItem.label}</span>
```
good to keep it so that we can provide support for custom design tokens

src/components/forms/FAQItem.scss
```diff
@@ -0,0 +1,17 @@
+@use '@utrecht/icon-css';
+
+@use '@/scss/bem';
+
+.openforms-faq-tooltip {
+  margin-block-start: var(--of-faq-tooltip-margin-block, 8px);
```
sergei-maertens: blocks/components are *not* allowed to set margins, because that affects the world outside of themselves

I'd to the route of a container element:

```scss
.openforms-faq-items {
  display: flex;
  flex-direction: column;
  row-gap: var(--of-faq-items-row-gap, 8px);
  padding-block-start: var(--of-faq-items-padding-block-start, 8px); // not sure if this is the best approach
}
```

This sets some default spacing, but allows theme designers to use different spacing values if they desire so.

src/components/forms/FAQTooltip.scss
```diff
@@ -0,0 +1,17 @@
...
+  padding-block-end: var(--of-faq-tooltip-padding-block, 11px);
+  padding-block-start: var(--of-faq-tooltip-padding-block, 11px);
+  padding-inline-end: var(--of-faq-tooltip-padding-inline, 16px);
+  padding-inline-start: calc(
+    var(--of-faq-tooltip-padding-inline, 16px) - var(--_border-left-width)
+  );
+
+  &__title:before {
+    content: ' ';
+  }
```
sergei-maertens: All this should not be necessary when you use the `.utrecht-link-button` component, because it already defines a column-gap token that creates the spacing between icon and label. That's also the right approach:

* it ensures accessibility
* it provides consistent spacing for button/link-like elements
* it can be themed through design tokens

src/components/forms/FAQTooltip.scss
```diff
@@ -0,0 +1,17 @@
...
+.openforms-faq-tooltip {
+  margin-block-start: var(--of-faq-tooltip-margin-block, 8px);
+  padding-block-end: var(--of-faq-tooltip-padding-block, 11px);
+  padding-block-start: var(--of-faq-tooltip-padding-block, 11px);
+  padding-inline-end: var(--of-faq-tooltip-padding-inline, 16px);
+  padding-inline-start: calc(
+    var(--of-faq-tooltip-padding-inline, 16px) - var(--_border-left-width)
+  );
+
+  &__title:before {
```
sergei-maertens: also, you should really be using the `@include bem.element('title')` sass selector here. You can then still include the pseudo element:

```scss
.openforms-faq-item {
  @include bem.element('title') {
    &:before { ... }
  }
}
```

sergei-maertens: I think we want to be prepared for some potential different typography through opt-in design tokens. If they're not defined, they'll just fall back to the defaults of the utrecht-link-button component (so use the link tokens), I believe:

```scss
.openforms-faq-item {
  @include bem.element('title') {
    color: var(--of-faq-item-title-color);
    font-size: var(--of-faq-item-title-font-size);
    font-weight: var(--of-faq-item-title-font-weight);
    line-height: var(--of-faq-item-title-line-height);
    text-decoration: var(--of-faq-item-title-text-decoration);
    
    &:hover {
      text-decoration: var(--of-faq-item-title-hover-text-decoration);
    }
  }
}
```
I don't think we need to define any of those design tokens ourselves at this point.

sergei-maertens (changes requested):
Please also improve you commit hygiene:

* The types version bump should be it's own atomic commit
* The implementation of the `FAQItems` component should be its own atomic feature commit
* Adding the feature to all components is its own feature commit too

I'm also missing more information in the commit message:

* why the types dependency needed bumping, and why it was moved from dependencies to dev dependencies
* why you opted for `div`s with `role="button"` instead of proper button elements
* why there are multiple text nodes for the same label of an FAQ item

src/components/forms/FAQItems.tsx
SonnyBA: See https://github.com/open-formulieren/types/pull/99 for the change in the types repository.

src/components/forms/Checkbox/Checkbox.tsx
```diff
@@ -1,10 +1,12 @@
+import type {FAQItem} from '@open-formulieren/types/dist/common';
```
sergei-maertens: TODO: once new version of types is available, update the import

src/components/forms/DateField/DateField.tsx
```diff
@@ -1,7 +1,9 @@
+import type {FAQItem} from '@open-formulieren/types/dist/common';
```
sergei-maertens: TODO: once new version of types is available, update the import

src/components/forms/DateTimeField/DateTimeField.tsx
```diff
@@ -1,3 +1,4 @@
+import type {FAQItem} from '@open-formulieren/types/dist/common';
```
sergei-maertens: TODO: once new version of types is available, update the import

src/components/forms/EditGrid/EditGrid.tsx
```diff
@@ -1,3 +1,4 @@
+import type {FAQItem} from '@open-formulieren/types/dist/common';
```
sergei-maertens: TODO: once new version of types is available, update the import

src/components/forms/InputGroup/InputGroup.tsx
sergei-maertens, line 1: Do we need the support for FAQ items in this component? I don't think so.

The problem here is that there's no slot for the validation errors, meaning validation errors going below the input group appear after the FAQ Items, which is opposite of what we want.

I think this component should not support the prop, and the components that use `InputGroup` are responsible of rendering the faq items themselves, after the validation errors display.

src/components/forms/FAQItems.tsx
```diff
@@ -0,0 +1,65 @@
...
+  );
+
+  const dialogOpenKeys = ['Enter', 'Space'];
+  return (
+    <>
+      <LinkButton
+        tabIndex={0}
+        className="openforms-faq-items__item"
+        onClick={() => setIsOpen(true)}
+        onKeyDown={(event: KeyboardEvent) => dialogOpenKeys.includes(event.code) && setIsOpen(true)}
```
sergei-maertens: does this not happen automatically through the `onClick`? I'd expect the native button element to cover that, once it has focus.

src/components/forms/FAQItems.tsx
```diff
@@ -0,0 +1,65 @@
...
+    <Modal title={label} variant="sheet" closeModal={() => setIsOpen(false)} isOpen={isOpen}>
+      <HTMLContent dangerouslySetInnerHTML={{__html: sanitizedContent}} />
+    </Modal>
+  );
+
+  const dialogOpenKeys = ['Enter', 'Space'];
+  return (
+    <>
+      <LinkButton
+        tabIndex={0}
```
sergei-maertens: When I remove this in the DOM in storybook, I can still tab properly, so I don't think this is necessary (`<button>` elements can be tabbed by default).

src/components/forms/FAQItems.tsx
```diff
@@ -0,0 +1,65 @@
...
+        className="openforms-faq-items__item"
```
sergei-maertens:
```suggestion
        className="openforms-faq-item"
```
the plural form `openforms-faq-items` is for the container, the singular form for a single item

src/components/forms/FAQItems.tsx
```diff
@@ -0,0 +1,65 @@
...
+        onClick={() => setIsOpen(true)}
+        onKeyDown={(event: KeyboardEvent) => dialogOpenKeys.includes(event.code) && setIsOpen(true)}
+      >
+        <UtrechtIcon>
+          <Icon icon="tooltip" />
+        </UtrechtIcon>
+
+        <span className="openforms-faq-items__title">{label}</span>
```
sergei-maertens:
```suggestion
        <span className="openforms-faq-item__title">{label}</span>
```
here too - use the singular form for proper BEM semantics

src/components/forms/FAQItems.tsx
```diff
@@ -0,0 +1,65 @@
...
+      </LinkButton>
+      {modalSheet}
+    </>
+  );
+};
+
+const FAQItems: React.FC<FAQItemsProps> = ({items = []}) => {
```
sergei-maertens:
```suggestion
const FAQItems: React.FC<FAQItemsProps> = ({items}) => {
```
default is not necssary, because the prop cannot be undefined/null from the types

src/components/forms/FAQItem.scss
```diff
@@ -0,0 +1,24 @@
...
+    font-size: var(--of-faq-item-title-font-size);
+    font-weight: var(--of-faq-item-title-font-weight);
+    line-height: var(--of-faq-item-title-line-height);
+    text-decoration: var(--of-faq-item-title-text-decoration);
+
+    &:hover {
+      text-decoration: var(--of-faq-item-title-hover-text-decoration);
+    }
+  }
+}
```
sergei-maertens, line 26:
```suggestion
.openforms-faq-items {
  display: flex;
  flex-direction: column;
  row-gap: var(--of-faq-items-row-gap, 8px);
  align-items: flex-start;
  padding-block-start: var(--of-faq-items-padding-block-start, 8px);
}

.openforms-faq-item {
  @include bem.element('title') {
    color: var(--of-faq-item-title-color);
    font-size: var(--of-faq-item-title-font-size);
    font-weight: var(--of-faq-item-title-font-weight);
    line-height: var(--of-faq-item-title-line-height);
    text-decoration: var(--of-faq-item-title-text-decoration);

    &:hover {
      text-decoration: var(--of-faq-item-title-hover-text-decoration);
    }
  }
}
```

Note that in your version you're mixing design tokens - a component with name `openforms-faq-items` cannot have a design token `of-faq-item` (singular) - the component names don't match!

The proper approach here is to just split in two components/blocks - one for the container, and one for an individual item.

sergei-maertens (changes requested):

src/components/forms/FAQItems.tsx
SonnyBA: It does :exploding_head: , TIL.

sergei-maertens (approved):

PR #337, opened 2026-06-15
Fix value display for times without seconds

### LINKED ISSUE #6356: Bug: Time component in Repeating Group shows "Invalid Date" after saving

3.5.3

### Customer reference

Taiga Maykin #248

Description
When using the Time Display component inside a Repeating Group, a valid time value is not shown after saving. Instead, the component displays the error message "Invalid Date".
Users are still able to proceed to the next step despite the "Invalid Date" .
Image

The time is correctly displayed in Overview and Submission
Image

Image

The same time component works correctly when used as a standalone field outside of a repeating group.

1. Open the form "time in repeating groups".
2. Add a new entry to the repeating group.
3. Enter a date and a time.
4. Click Save.

We want the time to be displayed, and not "Invalid Date"

### Which renderer?

None

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
Closes open-formulieren/open-forms#6356

The native time input (with minute-precision) passes the value as just HH:mm and omits the seconds part - which is valid according to the spec. Our time parts parsing code didn't take this possibility into account when preparing the value for display.

### DIFF
--- a/src/registry/time/ValueDisplay.tsx
+++ b/src/registry/time/ValueDisplay.tsx
@@ -16,14 +16,14 @@ const formatTime = (timeValue: string): React.ReactNode => {
   let time = new Date(timeValue);
 
   if (isNaN(Number(time))) {
-    const [hours, minutes, seconds] = timeValue.split(':');
+    const [hours, minutes, seconds = '00'] = timeValue.split(':');
     time = new Date();
     time.setHours(parseInt(hours));
     time.setMinutes(parseInt(minutes));
     time.setSeconds(parseInt(seconds));
   }
 
-  return <FormattedTime value={time} format="short" />;
+  return <FormattedTime value={time} />;
 };
 
 const ValueDisplay: React.FC<ValueDisplayProps> = ({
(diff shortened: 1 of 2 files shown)


src/registry/time/ValueDisplay.tsx
```diff
@@ -16,14 +16,14 @@ const formatTime = (timeValue: string): React.ReactNode => {
...
   if (isNaN(Number(time))) {
-    const [hours, minutes, seconds] = timeValue.split(':');
+    const [hours, minutes, seconds = '00'] = timeValue.split(':');
     time = new Date();
     time.setHours(parseInt(hours));
     time.setMinutes(parseInt(minutes));
     time.setSeconds(parseInt(seconds));
   }
 
-  return <FormattedTime value={time} format="short" />;
```
sergei-maertens, line 26: The `format=short` was leading to errors/warnings being emitted by react-intl, as there are no built-in formats for the time display (date does have them).

vaszig (approved):
Tested locally too, makes sense.

sergei-maertens:
Backported to stable/1.1.x in a7265a7848c45b9c05e7de3d78802385a999fa2a

PR #338, opened 2026-06-15
Ensure the first component with validation error(s) is scrolled into view

### LINKED ISSUE #6375: New renderer: Validation on submit does not scroll to top.

3.5.3

### Customer reference

Leiden 74

The new renderer seems to pick a random field to jump to. This used to be the "top" field in the step that produced the error/warning. 

old: 

Image

new:

Image

_No response_

_No response_

### Which renderer?

None

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
Closes open-formulieren/open-forms#6375

We use the formik submitCount to determine if we need to trigger a scroll action, to ensure this only happens after the user submits the form. We can't tap into Formik's validation chain, as the submit action triggers Formik validation, and only when the form is valid, onSubmit will be called, which is too late.

This mechanism queries the children of the form container for the first error element, based on the NL DS class names. We can't really test this easily, but it's expected the Chromatic snapshots will capture the scroll position properly. That's why the story uses 50 form fields.

### DIFF
--- a/src/components/FormioForm.stories.tsx
+++ b/src/components/FormioForm.stories.tsx
@@ -834,3 +834,33 @@ export const ValidationForMultipleDateComponents: Story = {
     expect(dateRequired).toHaveAttribute('aria-invalid', 'true');
   },
 };
+
+export const ScrollsFirstErrorIntoView: Story = {
+  args: {
+    components: [...Array(15).keys()].map(i => ({
+      id: `component-${i}`,
+      type: 'textfield',
+      key: `field-${i}`,
+      label: `Field ${i}`,
+      validate: {
+        required: false,
+        maxLength: 5,
+      },
+    })),
+  },
+
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+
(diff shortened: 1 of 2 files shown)


src/components/FormioForm.stories.tsx
```diff
@@ -834,3 +834,29 @@ export const ValidationForMultipleDateComponents: Story = {
...
+        required: false,
+        maxLength: 5,
+      },
+    })),
+  },
+
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+
+    const field10 = canvas.getByLabelText('Field 0');
```
vaszig: I guess you mean `field0` (but no worries of course). I would expect though to have a similar issue in `field10` for example (multiple errors), in order to see that indeed we scroll into the 1st component (which I tested and it's happening).

sergei-maertens: yeah I initially had 50 fields in the story to try to capture that in chromatic, but that failed horribly. Will update the variable name!

Good addition - I'll do the same for the 10th field :)

vaszig (approved):

PR #341, opened 2026-06-16
Fix excessive hooks firing without changed data

### LINKED ISSUE #6354: "latlng-search" and "address-autocomplete" are firing on every interaction

3.4.9 & 3.5.3

### Customer reference

DB: 162

This is seems to cause "throttling" for submitters, where forms have multiple location in steps, or a lot of questions after providing a address. 

Image

3.4.9: Seems to generate twice as much requests compared 3.5.3
3.5.3: only 1 request where you would normally see a "check_logic" call.

sergei-maertens:
Closes open-formulieren/open-forms#6354 (partly)

The coordinates prop is a non-primitive data structure, so recreating this in the render function of the map and passing it down to the NearestAddress component / useAsync hook will trigger changed dependencies, and invoke the location label lookup, triggering API calls.

Only when the values of the coordinates change should this be invoked, so we can enforce a stable identity with a strategic useMemo calls based on the root value.

### DIFF
--- a/src/components/map/LeafletMap.tsx
+++ b/src/components/map/LeafletMap.tsx
@@ -1,7 +1,7 @@
 import type {MapComponentSchema} from '@open-formulieren/types';
 import type {CoordinatePair, GeoJsonGeometry} from '@open-formulieren/types/dist/components/map';
 import L from 'leaflet';
-import {useEffect, useId, useRef} from 'react';
+import {useEffect, useId, useMemo, useRef} from 'react';
 import {useIntl} from 'react-intl';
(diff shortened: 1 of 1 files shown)


sergei-maertens:
W/r to the backport -> if this cherry picks cleanly to stable/1.1, we can backport. If not, the SDK fix for 3.4 with the cache would probably be sufficient.

vaszig (approved):

sergei-maertens:
Backported to stable/1.1.x in b936d41cb5c84302193a83ddb93c0ec31c04b019

PR #346, opened 2026-06-25
Remove backend compatibility code for processVisibility

### LINKED ISSUE #6276: Revert compatibility changes made to `processVisiblity` of the renderer
These were made to match the backend clearOnHide behavior. See `emulateBackend`, `clearValueCallback`, etc. of the evaluation context.

viktorvanwijk:
Closes open-formulieren/open-forms#6276

Burn it all! :fire:


viktorvanwijk:
Discussed during standup: we can actually leave this unmerged until right before the 4.0 release, to avoid dealing with stable branches for 3.5 backports. In the SDK, we just remove the `emulateBackend: true` flag to revert back to the default behaviour. 

I guess technically it is a breaking change, as we remove this emulate backend feature. Though one could argue whether this should have been in here in the first place :upside_down_face:

sergei-maertens:
> I guess technically it is a breaking change, as we remove this emulate backend feature. Though one could argue whether this should have been in here in the first place

It's not part of the [documented public API](https://open-formulieren.github.io/formio-renderer/?path=/docs/introduction--docs), so we can release it as minor version :-)

sergei-maertens (approved):

PR #347, opened 2026-07-01
Limit datepicker date to max date from validation

### LINKED ISSUE #6403: new renderer: Adjust datepicker to show/open on first available "pickable" date.
### Thema / Theme

Frontend

### Omschrijving / Description

Date fields in the new renderer always open on the current date(month).
When validation is applied to only allow dates that are:
`today` - `18 years` 
the user needs to click at least 18 times to be able to see and pick the first available date. 

This behaviour was not present in the old renderer. 

### Added value / Toegevoegde waarde

_No response_

### Aanvullende opmerkingen / Additional context

_No response_

sergei-maertens:
Closes open-formulieren/open-forms#6403

Ensure the users don't need to click tens of times when a maximum date(time) is configured on the date/datetime components.

### DIFF
--- a/src/components/forms/DateField/DatePicker/DatePicker.tsx
+++ b/src/components/forms/DateField/DatePicker/DatePicker.tsx
@@ -12,7 +12,7 @@ import Tooltip from '@/components/forms/Tooltip';
 
 import {useDateLocaleMeta} from '../hooks';
 import {PART_PLACEHOLDERS} from '../messages';
-import {parseDate} from '../utils';
+import {getBestInitialDate, parseDate} from '../utils';
 import './DatePicker.scss';
 
 interface DatePickerFieldProps {
@@ -181,7 +181,7 @@ const DatePickerField: React.FC<DatePickerFieldProps> = ({
                 await validateField(name);
                 setIsOpen(false);
               }}
-              currentDate={currentDate ?? undefined}
+              currentDate={currentDate ?? getBestInitialDate(minDate, maxDate)}
               minDate={minDate}
               maxDate={maxDate}
               displayWeekend={displayWeekend}
--- a/src/components/forms/DateTimeField/DateTimeField.tsx
+++ b/src/components/forms/DateTimeField/DateTimeField.tsx
@@ -16,7 +16,7 @@ import {useFieldConfig, useFieldError} from '@/hooks';
 import './DateTimeField.scss';
 import {useDateLocaleMeta} from './hooks';
 import {PART_PLACEHOLDERS} from './messages';
-import {parseDateTime} from './utils';
+import {getBestInitialDate, parseDateTime} from './utils';
 
 export interface DateTimeFieldProps {
   /**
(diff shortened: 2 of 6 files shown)


src/components/forms/DateField/DatePicker/DatePicker.tsx
```diff
@@ -181,7 +181,7 @@ const DatePickerField: React.FC<DatePickerFieldProps> = ({
                 await validateField(name);
                 setIsOpen(false);
               }}
-              currentDate={currentDate ?? undefined}
+              currentDate={currentDate ?? maxDate ?? undefined}
```
robinmolen: I don't know how this is/was implemented in the older renderer (i'll check it out), but the same issue happens with `minDate`.

When you set `minDate` in the future, like `2040-01-01`, the date picker presents the current date. So you will still have to click a bunch of times (and search for available dates) until you can fill in the input.

So i think the implementation should be `currentDate={currentDate ?? minDate ?? maxDate ?? undefined}`?
Selecting the first possible date if specified, with a fallback to the last possible date, and finally `now`?

robinmolen: Now that i'm thinking about it, maybe some "when is the max date?" logic is needed.

Lets say maxDate is set on `2040-01-01`, with no minDate. This means that the datepicker will be opened on january 2040, even though "today" is available. This would mean that the user still has to click ~14 times just to select this year.

So i think we need some "when is the first date" logic. Which looks for the first date in `maxDate`, `minDate` and `now`, and uses that as the default date? https://date-fns.org/v4.4.0/docs/min

src/components/forms/DateTimeField/DateTimeField.tsx
```diff
@@ -245,7 +245,7 @@ const DateTimeField: React.FC<DateTimeFieldProps> = ({
                 } as React.ChangeEvent<HTMLInputElement>;
                 onPartChange(e);
               }}
-              currentDate={currentDateTime ?? undefined}
+              currentDate={currentDateTime ?? maxDate ?? undefined}
```
robinmolen: The previous comments, about `minDate` and the comparison between today and `maxDate/minDate`, also apply here :)

robinmolen (changes requested):
I have some ideas to improve this

src/components/forms/DateField/DatePicker/DatePicker.tsx
sergei-maertens: The old renderer didn't explicitly do anything with this afaik, I think this was just some behaviour that was implemented in the flatpickr library.

I didn't consider min date in the future indeed, but that's a valid point and the suggestion holds up :-) will add stories + the patches!

sergei-maertens: Yeah good point, similarly to above I didn't consider max date in the future!

robinmolen (approved):

sergei-maertens:
Backported to stable/1.1.x in fd15dd44b0146683e63fd2bf86b051d91d684272

PR #348, opened 2026-07-01
Set up correct defaults for `ul` styling

### LINKED ISSUE #6410: Bullet points appearance change in last version

3.5.4

### Customer reference

#102

Bulletpoint appearance changed in 3.5.4, the bullets got bigger.  This seems independent from theme/stijl.

Image

Image

Image

Old version of OF still shows the right size:

Image

Image

1. Go to our test environment and check 'Bullet points bug' form

more balanced appearance of bullet points

### Which renderer?

_No response_

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

sergei-maertens:
Partly closes open-formulieren/open-forms#6410

Define fallback values that match the non-NL DS based components from earlier versions. See commit open-formulieren/design-tokens@004aed328d957690d856a7a18b58acda03b76a03 and open-formulieren/design-tokens@5e8d754e86886f6eb18bb28f435e52b9a2e85adc for more details. In a nutshell:

* font-size issues
* vertical alignment issues if the font-size gets fixed
* horizontal spacing off

### DIFF
--- a/src/registry/content/index.stories.tsx
+++ b/src/registry/content/index.stories.tsx
@@ -41,7 +41,18 @@ export const SimpleContent: Story = {
       cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non
       proident, sunt in culpa qui officia deserunt mollit anim id
       <a href="https://example.com" target="_blank" rel="external">est laborum</a>.</p>
-      <p>Second paragraph</p>`,
+      <p>Second paragraph</p>
+      <ul>
+        <li>Unordered list</li>
+        <li>
+          With nested items:
+          <ul>
+            <li>Item 1</li>
+            <li>Item 2</li>
+          </ul>
+        </li>
+      </ul>
+      `,
       customClass: '',
     } satisfies ContentComponentSchema,
   },
--- a/src/scss/shims.scss
+++ b/src/scss/shims.scss
@@ -5,10 +5,21 @@
 // have higher priority/selectivity.
 :root {
   --utrecht-form-control-font-weight: var(--of-input-font-weight, initial);
-  // adds the quotes that the default styles are missing
-  --utrecht-unordered-list-list-style-type: '●';
+
+  // defaults for the unordered list component to be backwards compatible. Can be
+  // removed in Open Forms 5.0. (DeprecationWarning)
+  --utrecht-unordered-list-list-style-type: disc;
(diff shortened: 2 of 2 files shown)


viktorvanwijk (commented):
So for my understanding, we are reverting to how it looked before 3.5, and will be using the new styling from 5.0?

Also, it seems counter intuitive to me that changing design tokens for the unordered list has an effect on the ordered list as well?

sergei-maertens:
> So for my understanding, we are reverting to how it looked before 3.5, and will be using the new styling from 5.0?
> 
> Also, it seems counter intuitive to me that changing design tokens for the unordered list has an effect on the ordered list as well?

3.5.4 (or .3?) patched the font-size for the customer profile and used the opportunity to replace our custom CSS that doesn't support design tokens with the Utrecht `ordered-list` NL DS component that does support theming. This required some patching to make it all work properly and retain the same list markers - level 0 is bullets/discs, level 1 is dashes. However, there was a regression there by setting `list-style-type` to a bullet utf-8 character rather than the value `disc` (browser default + the style as it has always been in our theme), which led to too-big font-sizes for the marker + the horizontal spacing being messed up.

So, this patch addresses that regression: it restores the small disc marker and the spacing is restored, fixing the alignment issues. Unless we get a redesign of our theme in this aspect, this is the style that we'll keep, no changes in 5.0, but in 5.0 we'll drop the CSS shim and then the design tokens must be explicitly specified in our own theme (they are now with the new PR) and custom themes (this we don't control and need a deprecation period to not force late stage breaking changes in themes).

> Also, it seems counter intuitive to me that changing design tokens for the unordered list has an effect on the ordered list as well?

That caught me off guard as well! Apparently if the ordered list design tokens are not defined, they fall back to the unordered list tokens. I understand why they made it like that, but it breaks the whole components-are-isolated-and-independent principle (@robbert are you reading along? :D ). Luckily Chromatic caught the visual regressions - I've amended my commit to address/fix that, because the lack of spacing looks cluttered imo.

viktorvanwijk:
Thanks for the explanation :)

> I've amended my commit to address/fix that, because the lack of spacing looks cluttered imo.

Nice, I agree!

viktorvanwijk (approved):

sergei-maertens:
Backported to stable/1.1.x in 78dddc8650169c671000b0487cb4935b3ac9e6e1

PR #349, opened 2026-07-02
Track parent-child relations of components

viktorvanwijk:
Necessary for the SDK to be able to look up the parent to inspect the hidden state when evaluating logic rules.

Was initially added, but then removed as the backend didn't need this when processing clearOnHide. Now that the behavior has been reworked, we actually do need it :)

(cherry picked from commit cc716d56c78166fd80cec4f95e7b4511755a95d2)


sergei-maertens (approved):
Code looks familiar 😉

PR #352, opened 2026-07-13
Fix uploads in repeating group

### LINKED ISSUE #6423: New renderer: uploads in repeating group missing in data

3.4.x + 3.5.x

### Customer reference

DH 883

The files uploaded in a repeating group are missing from the submission data. It's not possible to save the group, nor navigate to the next step. 

Image

files are present in the submission

_No response_

_No response_

### Which renderer?

None

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

viktorvanwijk:
Closes open-formulieren/open-forms#6423

### DIFF
--- a/src/registry/file/File.tsx
+++ b/src/registry/file/File.tsx
@@ -9,6 +9,7 @@ import HelpText from '@/components/forms/HelpText';
 import Label from '@/components/forms/Label';
 import Tooltip from '@/components/forms/Tooltip';
 import ValidationErrors from '@/components/forms/ValidationErrors';
+import {useFieldConfig} from '@/hooks';
 
 import './File.scss';
 import UploadInput from './UploadInput';
@@ -23,7 +24,8 @@ export interface FormioFileProps {
 }
 
 const FormioFile: React.FC<FormioFileProps> = ({componentDefinition}) => {
-  const {key: name} = componentDefinition;
+  const {key} = componentDefinition;
+  const name = useFieldConfig(key);
   return (
     <FieldArray name={name} validateOnChange={false}>
       {arrayHelpers => (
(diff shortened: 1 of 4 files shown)


src/registry/editgrid/index.stories.ts
```diff
@@ -769,6 +770,64 @@ export const NestedWithSimpleConditionalsAndInitialErrors: Story = {
...
+      } satisfies FormSettings['componentParameters'],
+    },
+    formik: {
+      initialValues: {
+        editgrid: [],
+      },
+    },
+  },
+  play: async () => {
+    // TODO-6423: what to write here?
```
sergei-maertens: Simulate a file upload, submit the form and assert on the onSubmit mock that the expected data is there?

I'd write this as a regression/unit test rather than a story though, in some .spec.tsx file

src/registry/file/File.tsx
```diff
@@ -23,7 +24,8 @@ export interface FormioFileProps {
 }
 
 const FormioFile: React.FC<FormioFileProps> = ({componentDefinition}) => {
-  const {key: name} = componentDefinition;
+  const {key} = componentDefinition;
+  const name = useFieldConfig(key);
```
sergei-maertens, line 28: D'oh 🤦‍♀️

sergei-maertens (commented):
At least the fix looks simple enough 😅

src/registry/editgrid/index.stories.ts
viktorvanwijk: Hmm OK, I will have a look. At first was thinking about interacting with the upload component, but I see there are no existing tests/stories for that. Probably for a good reason, because it looks annoying :sweat_smile:

viktorvanwijk: After some pain, it's working :tada:

sergei-maertens (approved):

viktorvanwijk:
Backport to 1.1.x: 360b7a7c

PR #353, opened 2026-07-14
Fix dynamic options value type in Formik state

### LINKED ISSUE #6420: Dynamic items added with logic in Select Boxes act strange
[strangeon.zip](https://github.com/user-attachments/files/29497086/strangeon.zip)

3.5.x

### Customer reference

DH 880

When selecting a dynamic item in a Select Boxes component - the value isn't a boolean but instead a ['on'] is set as value:

```
{
    "a": false,
    "b": [
        "on"
    ]
}
```
which crashes the form. 
The Select boxes component needs to be loaded with some item (a), and with logic in the step change to include more items (b). the new items shows this behaviour while the initial item acts like a boolean. 

_No response_

_No response_

### Which renderer?

_No response_

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

sergei-maertens:
Closes open-formulieren/open-forms#6420

Ensure that the Formik field(s) state is in sync with the available options for selectboxes.

This is a problem only for the selectboxes component because it has a weird datastructure of `Record<string, boolean>`, which leads to a Formik field leaf node for each option on the component. Radio & select don't have that issue, because their option names don't end up in the field names.

The fix ensures that at all times the formik field names match the option values defined on the component.

**TODO**

- [x] Test if selectboxes work in editgrids, the same bug as for file uploads may be present
- [x] Test that option removal leads to it not being in the `onSubmit` values

### DIFF
--- a/src/components/FormioForm.spec.tsx
+++ b/src/components/FormioForm.spec.tsx
@@ -1204,4 +1204,51 @@ describe('Regressions', () => {
     await expect.poll(() => onSubmit).toHaveBeenCalledOnce();
     expect(onSubmit).toHaveBeenCalledWith({selectboxes: {b: true}});
   });
+
+  // Suspected bug similar to open-formulieren/open-forms#6423
+  test('Selectboxes inside editgrid', async () => {
+    const onSubmit = vi.fn();
+    const screen = await render(
+      <Form
+        components={[
+          {
+            type: 'editgrid',
+            id: 'editgrid',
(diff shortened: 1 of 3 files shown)


robinmolen (approved):
Looks good!

sergei-maertens:
Backported to stable/1.1.x in 7fb0236b8166258729e532e92e766108549df567

PR #362, opened 2026-07-20
Fix some more editgrid bugs

### LINKED ISSUE #357: Infinite render loop when using a (selectboxes) conditional in editgrids
When you have a conditionally displayed field inside an edit grid item that depends on the value of a sibling, especially with selectboxes, the UI becomes unresponsive to checkbox clicks. Opening the debugger reveals that it ended up in an infinite render cycle:

Image

sergei-maertens:
Closes #357

Fixes three separate bugs related to editgrids:

* prevent infinite render loop when using a selectboxes conditional in an editgrid, due to missing `componentsMap` entries
* prevent visibility processing & value updates from being committed to the editgrid values despite not having been saved by the user
* ensure "touched" state is correctly determined inside edit grid by using the properly prefixed field name


viktorvanwijk (approved):
Nice, thanks for the clear commit messages :)

sergei-maertens:
Backport commits to stable/1.1.x:

* d268155de742d3c436c5e3b21e6f57f973ff1fda
* eca53d7060c92bca9bb19e01c2fb30fce4ca29a9
* 23d8ea3443537e68adf486e89914b6c9b14527a8

The regression tests are not backported (except for one) because they use vitest-browser mode which isn't available in 1.1.x

PR #363, opened 2026-07-23
Update validation error positioning & error styling

### LINKED ISSUE #6182: Validation error positioning
NL DS leans towards putting validation errors above the field (consistent with the gov.uk approach), while the Open Forms theme puts them below. We want to avoid users having to override the container/wrapper component layout just to achieve this, as there should be CSS-only solutions.

Possible options:

* flex-direction row + the order property
* CSS grid with named slots/subgrids

sergei-maertens:
Closes open-formulieren/open-forms#6182

---

:boom: Before merging this, we need to create a new `stable/1.y.x` branch for bugfix backports. This will effectively introduce renderer 2.0 due to the big visual changes.


sergei-maertens:
Marking RFR despite it not being so just to get the chromatic snapshot comparisons :)

src/components/forms/DateField/DateInputGroup/DatePartInput.tsx
```diff
@@ -1,9 +1,9 @@
-import {FormLabel} from '@utrecht/form-label-react';
 import {Textbox} from '@utrecht/textbox-react';
 import {useId} from 'react';
 import {FormattedMessage, defineMessages, useIntl} from 'react-intl';
 import type {MessageDescriptor} from 'react-intl';
 
+import {InputGroupItemLabel} from '../../InputGroup';
```
sergei-maertens: Fix import to not be relative

src/components/forms/EditGrid/EditGridItem.tsx
```diff
@@ -180,15 +180,13 @@ function EditGridItem<T extends {[K in keyof T]: JSONValue} = JSONObject>({
             }}
           >
             <>
-              {props.getBody({expanded: isExpanded})}
-
               {itemError && itemErrorId && (
                 // wrapper div to workaround `order` CSS from utrecht-form-field
```
sergei-maertens: obsolete comment

src/components/forms/TextField/TextField.tsx
```diff
@@ -80,13 +80,22 @@ export interface TextFieldProps {
...
+  afterInput?: React.ReactNode;
 }
 
 /**
  * A plain textfield form field, complete with label, help text and validation errors.
  *
  * The form field state is managed with Formik, but this should only be an
  * implementation detail within this library.
+ *
+ * @todo Refactor this to `FormFieldTextbox` or whatever the NL DS component will be,
```
sergei-maertens, line 96: not for now :)

src/components/forms/Fieldset.scss
```diff
@@ -1,76 +1,62 @@
...
+      }
 
-    @include bem.modifier('tooltip') {
-      // See src/components/forms/Tooltip.scss - over time this custom component will be
-      // replaced with the utrecht-fieldset component.
-      display: flex;
-      align-items: var(--of-utrecht-form-field-label-tooltip-align-items, center);
-      column-gap: var(--of-utrecht-form-field-label-tooltip-column-gap, 4px);
+      @include bem.modifier('with-border') {
+        display: flex;
```
sergei-maertens, line 18: can't believe this actually works 🤯

src/components/forms/DateField/DateInputGroup/DateInputGroup.tsx
```diff
@@ -126,30 +137,34 @@ const DateInputGroup: React.FC<DateInputGroupProps> = ({
...
-      tooltip={tooltip}
-      isRequired={isRequired}
-      isReadOnly={isReadOnly}
-      name={name}
-    >
-      <DateInputItems
-        year={year}
-        month={month}
-        day={day}
+    <div>
```
SonnyBA, line 140: Is this div needed for anything?

src/components/forms/DateTimeField/DateTimeField.tsx
```diff
@@ -211,39 +211,54 @@ const DateTimeField: React.FC<DateTimeFieldProps> = ({
...
-                placeholder={placeholder}
-                autoComplete="off"
-              />
-              <DatePickerTrigger
-                className="openforms-datepicker-textbox__calendar-toggle"
-                disabled={isReadOnly}
-              />
-            </Paragraph>
+            <div
+              className={clsx('utrecht-form-field__input')}
```
SonnyBA, line 229: This seems like a confusing class name to use for a container element 🤔

src/registry/addressNL/index.tsx
```diff
@@ -96,6 +96,15 @@ export const FormioAddressNL: React.FC<FormioAddressNLProps> = ({
       })}
       aria-describedby={[descriptionid, errorMessageId].filter(Boolean).join(' ') || undefined}
     >
+      {/*
+        no point wrapping these as the __description and __error-message elements don't
```
SonnyBA, line 100: Should we add a TODO here (to make it easier to find when searching for TODO's)?

src/components/forms/DateField/DateInputGroup/DateInputGroup.tsx
sergei-maertens: Yeah! Without it, the multi field delete button would be appended to the inputs instead of being isolated to the right

src/components/forms/DateTimeField/DateTimeField.tsx
sergei-maertens: Ha yeah, it's a class name for a css grid slot (from the NL Design System component). It basically says the input element(s) are contained in there somewhere

src/registry/addressNL/index.tsx
sergei-maertens: I'm not aware of any planning of these being added - if the fieldset component gets updated though, we'll have to update all uses of it so I think we'll find it quite easily.

